PR #1, opened 2021-11-15
Update script.js: formatted dislike counter

bstka:
Before this dislike counter show raw fetched data, eg: `10000` -> `10000`.
image

After little tweaking it shows how should it be, eg: `10000` -> `10K`
image


Anarios:
Thanks!

PR #4, opened 2021-11-15
Add UserScript

JRWR:
Tested in Brave using TamperMonkey, this script will allow easier install even for newbies as tampermonkey, grease monkey is on the store.


Anarios (approved):
Nice, thanks

PR #19, opened 2021-11-19
Added stats to the readme

DARKDRAGON532:


SirLich:
Technically "would be" was correct. 

When speaking about the future in the past, "would" is correct. 

"He will sing" (future)
"He did sing" (past)
"He would sing" (past, referencing future event)

Anarios (approved):

PR #25, opened 2021-11-21
Add Greasemonkey support

dshoreman:
First, thanks for creating this script/extension!

I installed the Userscript in Greasemonkey but, while the script ran, it would stop at the call to `GM_xmlhttpRequest` which apparently isn't available in GM.

Changing it to `GM.xmlHttpRequest(...)` got everything working great (after adding some logging to track the issue down, so I've included that to help find future glitches) so this PR adds it as a fallback in case the `GM_` variant can't be found.

As far as I can tell, `GM_` is available in both Tampermonkey and Violentmonkey. Although they seem to alias the `GM.` variant, I've kept both in case that's missing in some versions or other compatible plugins I'm unaware of.

Tried in both Firefox and Chrome on Linux with {Grease,Tamper,Violent}Monkey.

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -15,6 +15,11 @@
 // @updateURL https://github.com/Anarios/return-youtube-dislike/raw/main/Extensions/UserScript/Return%20Youtube%20Dislike.user.js
 // @grant GM_xmlhttpRequest
 // ==/UserScript==
+function cLog(text, subtext = '') {
+  subtext = subtext.trim() === '' ? '' : `(${subtext})`;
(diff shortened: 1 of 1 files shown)


Anarios:
Can you look at lates code please, before your changes? We just changed it to GM.xmlHttpRequest as well

Anarios (approved):

Anarios:
Merged, lots of other good improvements as well, thanks.

dshoreman:
Ah my apologies, saw the recent commits but didn't notice the change to the dot version 

I've also added a native fallback (keeping the extension-provided ones since they might provide extra functionality) so I'll create a separate PR for that shortly. Didn't realise you merged already! Thanks :)

PR #12, opened 2021-11-19
[Userscripts] remove dependency on Tampermonkey's function

wangchou:
**GM_xmlhttpRequest** is used in current UserScripts.
That function is only supported by Tampermonkey (price: $1.99 in Mac App Store).

New script is tested in Safari with free [UserScripts](https://apps.apple.com/us/app/userscripts/id1463298887?mt=12) extension for Mac App Store.

PR #26, opened 2021-11-21
Add native XMLHttpRequest fallback

dshoreman:
This implements part of #12, but maintaining the use of the `GM` functions where available. Instead it reuses the same options, simply wrapping the response in a plain object to avoid rewriting too much.


Anarios (approved):
Thanks a lot!

PR #30, opened 2021-11-22
Added a bug issue template

DARKDRAGON532:
image
This is how it looks like.
If you like it I'll also add one for feature suggestion etc.


Anarios:
Thanks a lot!

PR #31, opened 2021-11-22
Added chrome web store stats to readme

DARKDRAGON532:


Anarios:
I would wait with it untill stats are updated, it's stuck at 0 reviews right now.

Anarios (approved):

PR #42, opened 2021-11-23
Make like/dislike bar and count tooltip look exactly same as youtube originally had

yashptel:
- made like/dislike bar look exactly same as youtubes'
- also used youtube's color variables so we don't have to worry about changing the color for dark and light theme
- also made area for hovering to show like/dislike count larger(same as it was in youtube)
- used youtube's tooltip instead of making our own. so it matches the youtube's theme seemlessly
- I have made the changes for both chrome and firefox

Note: I have only tested my changes in chrome and edge chromium. So Please test the firefox version before merging.

Preview:
[https://imgur.com/a/uB5LAAa](https://imgur.com/a/uB5LAAa)

### DIFF
--- a/Extensions/chrome/content-style.css
+++ b/Extensions/chrome/content-style.css
@@ -1,46 +1,26 @@
 #return-youtube-dislike-bar-container {
-  background: #c6c6c6;
+  background: var(--yt-spec-icon-disabled);
+  border-radius: 2px;
 }
 
 #return-youtube-dislike-bar {
-  background: #5094c8;
-}
-
-[dark] #return-youtube-dislike-bar-container {
-}
-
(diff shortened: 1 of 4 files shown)


Anarios (changes requested):
Please don't don't apply styling to code that's not changed. Or apply default prettier codestyle.

yashptel:
> Please don't don't apply styling to code that's not changed. Or apply default prettier codestyle.

I have not. I have saved the file without formatting.

Anarios:
> > Please don't don't apply styling to code that's not changed. Or apply default prettier codestyle.
> 
> I have not. I have saved the file without formatting.

there are spaces changes, etc. But okay, I'll not waste your time and I'll format it myself.

Thanks a lot for your help

PR #43, opened 2021-11-23
Change Youtube to YouTube

trybick:
Change the extension titles from "Return Youtube Dislike" to "Return **YouTube** Dislike". Also updated a few other places in the code.


Anarios (approved):
Thanks!

PR #54, opened 2021-11-24
Tampermonkey // Userscript Optimizations

Frontesque:
simple stuff. Doesn't need to be done, but uses less resources for the same basic tasks

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -21,11 +21,8 @@ function cLog(text, subtext = '') {
 }
 
 function doXHR(opts) {
-  if (typeof GM_xmlhttpRequest === 'function') {
+  if (typeof GM_xmlhttpRequest === 'function' || typeof GM.xmlHttpRequest === 'function') {
     return GM_xmlhttpRequest(opts);
-  }
-  if (typeof GM.xmlHttpRequest === 'function') {
-    return GM.xmlHttpRequest(opts);
   }
 
   console.warn('Unable to detect UserScript plugin, falling back to native XHR.');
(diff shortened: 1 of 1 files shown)


Extensions/UserScript/Return Youtube Dislike.user.js
```diff
@@ -21,12 +21,9 @@ function cLog(text, subtext = '') {
 }
 
 function doXHR(opts) {
-  if (typeof GM_xmlhttpRequest === 'function') {
+  if (typeof GM_xmlhttpRequest === 'function' || typeof GM.xmlHttpRequest === 'function') {
```
SuperStormer, line 24: This is not equivalent to the previous code - if `GM.xmlHttpRequest` exists, but not `GM_xmlhttpRequest`, your commit will error, while the previous code won't.

Extensions/UserScript/Return Youtube Dislike.user.js
```diff
@@ -73,11 +70,9 @@ function isVideoNotDisliked() {
 function getState() {
   if (isVideoLiked()) {
     return "liked";
-  }
-  if (isVideoDisliked()) {
+  } else if (isVideoDisliked()) {
```
SuperStormer, line 73: Doesn't do anything since the first if statement's body always returns.

Extensions/UserScript/Return Youtube Dislike.user.js
Frontesque: I tested it and its what im currently running. It's working perfectly fine

SuperStormer: Yes, for your specific case it *happens* to work. But your change breaks other cases.

SuperStormer:
Note that these "optimizations" barely affect performance, if they even affect it all.

Extensions/UserScript/Return Youtube Dislike.user.js
dshoreman: @Frontesque To clarify why it didn't break anything for you, the code removed isn't for Tampermonkey.

Tampermonkey and (some?) other forks seem to use the `GM_`-prefixed function, but GreaseMonkey (Firefox) *only* has the `GM` class so both variations are necessary.

PR #55, opened 2021-11-24
Improve readability and remove unused functions

hwgilbert16:
- Moved an extremely long else-if train to switch instead of readability. 
- Removed unused/commented out code that appears to be from previous debugging that did not need to be included in a main branch.

### DIFF
--- a/Extensions/chrome/background.js
+++ b/Extensions/chrome/background.js
@@ -9,62 +9,51 @@ const apiUrl = "https://return-youtube-dislike-api.azurewebsites.net";
 
 chrome.runtime.onMessageExternal.addListener(
   (request, sender, sendResponse) => {
-    if (request.message === "get_auth_token") {
-      // chrome.identity.getAuthToken({ interactive: true }, function (token) {
-      //   console.log(token);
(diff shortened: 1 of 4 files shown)


Extensions/chrome/background.js
```diff
@@ -9,62 +9,51 @@ const apiUrl = "https://return-youtube-dislike-api.azurewebsites.net";
...
+              "Content-Type": "application/json",
+            },
+            body: JSON.stringify(toSend),
           });
-          sendResponse(result);
-        });
+          for (const toSendUrl of toSend) {
+            sentIds.add(toSendUrl);
+          }
+          toSend = [];
```
SuperStormer, line 41: missing a `break` here.

Extensions/firefox/background.js
```diff
@@ -1,66 +1,55 @@
...
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify(toSend),
+        });
+        for (const toSendUrl of toSend) {
+          sentIds.add(toSendUrl);
+        }
+        toSend = [];
```
SuperStormer, line 35: and here

PR #57, opened 2021-11-25
Narrow scope of `@match` clause

SuperStormer:
This restricts the userscript to only load on `*://*.youtube.com/watch*`, which prevents it from loading on pages where it is unnecessary.


j14-c:
Should be changed to `*://*.youtube.com/*` or it won't work when clicking a video from YouTube's homepage.

PR #61, opened 2021-11-26
Get dislike count directly from Youtube instead in UserScript

teppyboy:
## Why?
Because the current UserScript fetch dislikes count from https://return-youtube-dislike-api.azurewebsites.net/, which is not very accurate (and the youtube dislikes count in their API is still alive until Dec 13th), so for now we can get directly from Youtube to give the user a more accurate dislike count
## How?
Since I don't know **any** of JavaScript, I just copied code from https://github.com/Anarios/return-youtube-dislike/blob/main/Extensions/chrome/return-youtube-dislike.background.js#L56, and you're free to edit the code in this PR

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -15,159 +15,201 @@
 // @grant GM.xmlHttpRequest
 // ==/UserScript==
 function cLog(text, subtext = '') {
-  subtext = subtext.trim() === '' ? '' : `(${subtext})`;
-  console.log(`[Return Youtube Dislikes] ${text} ${subtext}`);
+    subtext = subtext.trim() === '' ? '' : `(${subtext})`;
+    console.log(`[Return Youtube Dislikes] ${text} ${subtext}`);
 }
 
+/*
 function doXHR(opts) {
-  if (typeof GM_xmlhttpRequest === 'function') {
+if (typeof GM_xmlhttpRequest === 'function') {
     return GM_xmlhttpRequest(opts);
-  }
-  if (typeof GM.xmlHttpRequest === 'function') {
-    return GM.xmlHttpRequest(opts);
-  }
-
-  console.warn('Unable to detect UserScript plugin, falling back to native XHR.');
-
-  const xhr = new XMLHttpRequest();
-
-  xhr.open(opts.method, opts.url, true);
-  xhr.onload = () => opts.onload({
-    response: JSON.parse(xhr.responseText),
-  });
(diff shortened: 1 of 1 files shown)


DARKDRAGON532:
The YouTube API requires an API key to use, the API key has a quota of 10,000 requests per day, and getting more quota is very hard, and probably won't be given to us since we are modifying the youtube ui.
Meanwhile the API we are using is maintained and coded by the owner of this extension.
We cant exceed the quota so we have to optimise.
> 3. How does this work? 
> 
> The extension collects the video id of the video you are watching, fetches the dislike (and other fields like views, likes etc) using our API, if this is the first time the video was fetched by our API, it will use the YouTube API to get the data, then stores the data in a database for caching (cached for around 2-3 days) and archiving purposes and returns it to you. The extension then displays the dislikes to you.
> 4. What will happen after the YouTube API stops returning the dislike count?
> 
> The backend will switch to using a combination of archived dislike stats, estimates extrapolated from extension user data and estimates based on view/like ratios for videos whose dislikes weren't archived and for outdated dislike archives.

If you really want accurate data, there are some other extensions that fetch directly from the YouTube API (you have to provide your own API key), but those extensions wont last after the api removes the dislike field.

teppyboy:
I believe the code to get YouTube dislikes count (at least in this project Chrome extension) doesn't use YT API, it just extract info from the HTML page, but yeah extensions won't last after youtube remove the dislike field.

DARKDRAGON532:
> I believe the code to get YouTube dislikes count (at least in this project Chrome extension) doesn't use YT API, it just extract info from the HTML page, but yeah extensions won't last after youtube remove the dislike field.

The backend for this extension uses the API, and it will even last after youtube removes the dislike field.

teppyboy:
Yeah, I know that this extension use the API, but [this line](https://github.com/Anarios/return-youtube-dislike/blob/main/Extensions/chrome/return-youtube-dislike.background.js#L56) says the extension can still fetch the info from YT, so what I'm trying to do here is use that code into the UserScript (because the backend right now is not very accurate)

DARKDRAGON532:
> Yeah, I know that this extension use the API, but [this line](https://github.com/Anarios/return-youtube-dislike/blob/main/Extensions/chrome/return-youtube-dislike.background.js#L56) says the extension can still fetch the info from YT, so what I'm trying to do here is use that code into the UserScript (because the backend right now is not very accurate)

My bad, I overlooked that.
If you cannot see the dislikes normally I doubt you would be able to fetch the link and get the dislike from there, anyway, the code seems fine, if its working for you then continue using it but I doubt it will go into the main branch.

PR #67, opened 2021-11-27
ESLint integration

DmACKGL:
Its always recommended to use ESLint when is possible.


Website/pages/links.vue
```diff
@@ -1,66 +1,63 @@
...
+    <section style="position: relative">
       <div class="center">
         <center>
+          <h1 style="font-size: 3em; margin-bottom: 0">Project Links</h1>
 
-          <h1 style="font-size: 3em; margin-bottom: 0;">Project Links</h1>
-          
           <div style="color: #999">
-            <p style="margin-bottom: 1em;">Links to the project and it's developers</p>
+            <p style="margin-bottom: 1em">
```
gitdev-bash, line 9: Does this break something?

gitdev-bash (approved):
Can verify everything but the dependency file ```Website/package-lock.json```.
Am not sure if ```Website/pages/links.vue``` line 9 breaks something.

Website/pages/links.vue
PickleNik: I don't see why it would, just moves it to the next line.
Though it does create conflicts for merge.
It would be ideal to leave the config files in and formatted files out.

PickleNik:
> Can verify everything but the dependency file `Website/package-lock.json`. Am not sure if `Website/pages/links.vue` line 9 breaks something.

Why do you think it would break something?

PickleNik (changes requested):
Could you remove formatted `.vue` files from commit? Or pull recent and re-format those. (to avoid conflicts)
Config files seem fine to me 👍🏼

Kefta:
Contains unnecessary changes and semantical edits.

PickleNik (approved):

gitdev-bash:
> Contains unnecessary changes and semantical edits.

Is a Lint unnecessary?

PickleNik:
This has been hanging for a couple weeks, so I re-implemented it in #280 , thanks for the suggestion tho

PR #79, opened 2021-11-28
Fix for locales separating decimals with commas

francesco-ferraro:
Some locales use a comma to separate decimals instead of a period. In these locales the number of dislikes is incorrectly shown:
image

This fix adds a simple regex to a str.replace


Anarios (approved):
Great, thanks!

j14-c:
@francesco-ferraro You forgot to escape the `.`

```js
'10560.0'.replace(/.0|,0/gi, ''); // "5"
'10560.0'.replace(/\.0|,0/, ''); // "10560"
```

PR #119, opened 2021-11-30
Bug Fix: Duplicate keys in v-for cause vue errors.

mattphelps-8451:
See here: https://stackoverflow.com/questions/51086657/vue-warn-duplicate-keys-detected-x-this-may-cause-an-update-error/52658488

Picture of errors before fix
Screen Shot 2021-11-30 at 1 07 51 PM
:


mattphelps-8451:
Just beaten by #122

PR #131, opened 2021-12-01
Enhancement: Add Opera/Edge links to Install Page

mattphelps-8451:
Many users are unaware they can enable the Chrome web store extension on their Chromium browsers. This PR adds icons, links, and reformatting. The new Opera and Edge links point to the same Chrome web store but in a more intuitive manner for other Chromium browser users by indicating they work for Edge and Opera. 

Note: I couldn't find a material design icon for Brave (unfortunately it's not native to vuetify and would need to be included in a different icon set).

Here's the new install page. TL;DR The new Edge/Opera links go to compatible Chrome web store.
Screen Shot 2021-12-01 at 1 28 20 AM

### DIFF
--- a/Website/pages/install.vue
+++ b/Website/pages/install.vue
@@ -5,8 +5,13 @@
 
     <div style="color: #999">
       <p style="margin-bottom: 0;"> This is an <b>ALPHA version!</b> It may be slow. It may be buggy.</p>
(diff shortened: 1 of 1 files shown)


Anarios (approved):
Thanks a lot!

Issue #142, opened 2021-12-01
(Bug): Returning NAN instead of 0 for dislikes


KJ002:
### Browser

Firefox

### Browser Version

Mozilla Firefox 94.0

### Extension or Userscript?

Extension

### Extension/Userscript Version

1.0.0.2

### Video link where you see the problem

https://www.youtube.com/watch?v=CP4jHWSdmpU

### What happened?

The dislikes displayed `NAN` where it should display `0`.
![nan](https://user-images.githubusercontent.com/36408549/144327432-b645ed39-ce14-4197-bb08-241a42c24ab2.png)

Paulmski:
Hi there, I was unable to reproduce with the following:
Fedora 35
Firefox
Mozilla Firefox 94.0
Extension 1.0.0.2

Video:
[https://www.youtube.com/watch?v=lz8-fdVtox8]()

Result:

![Screenshot from 2021-12-01 19-23-27](https://user-images.githubusercontent.com/78445698/144335761-44de14e4-f43f-4aa3-90db-36db1317faff.png)


Do you have any other information that might help reproduce? Does this happen with other videos?

TheJuki:
https://www.youtube.com/watch?v=xAlQ-UF6xgw

I saw it on this video. Refreshing showed 3 dislikes. I think because I was the only one to watch with the extension first.

Paulmski:
Thanks! I can reproduce this, seems to be as you say. Only videos that have never been queried for have this behavior, or so it seems.

samir-dahal:
unable to reproduce
![image](https://user-images.githubusercontent.com/39849021/144442471-df95136a-1052-4305-a106-33a438437eb9.png)

Quendi6:
I am on Firefox, while the page is loading I see the dislikes count, until the page has finished loading. Then I see NAN, for each video.

Paulmski:
> [return youtube dislike]: response from api:
return-youtube-dislike.script.js:7 [return youtube dislike]: {"type":"https://tools.ietf.org/html/rfc7231#section-6.5.4","title":"Not Found","status":404,"traceId":"00-823929c12977591ace4e4d7db96a7c04-a9f4e8aa50e8ea0d-00"}
watch:1 Error handling response: TypeError: Cannot read properties of undefined (reading 'toLocaleString')
at createRateBar (chrome-extension://gebbhagfogifgggkldgodflihgfeippi/return-youtube-dislike.script.js:198:44)
[...]

PR #153, opened 2021-12-02
Bug fix(Issue #142)

LaszloRobert:
Issue : Returning NAN instead of 0 for dislikes #142 
Changes:  `response.likes` or  `response.dislikes` if it is 0 then it will be interpreted as `false` and it won't reach the code to display the correct format.

### DIFF
--- a/Extensions/chrome/return-youtube-dislike.script.js
+++ b/Extensions/chrome/return-youtube-dislike.script.js
@@ -75,7 +75,7 @@
           cLog("response from youtube:");
           cLog(JSON.stringify(response));
           try {
-            if (response.likes || response.dislikes) {
+            if (response.likes >=0 || response.dislikes >=0) {
               const formattedDislike = numberFormat(response.dislikes);
               setDislikes(formattedDislike);
               createRateBar(response.likes, response.dislikes);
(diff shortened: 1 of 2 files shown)


Extensions/chrome/return-youtube-dislike.script.js
```diff
@@ -75,7 +75,7 @@
           cLog("response from youtube:");
           cLog(JSON.stringify(response));
           try {
-            if (response.likes || response.dislikes) {
+            if (response.likes >=0 || response.dislikes >=0) {
```
SukkaW, line 78: Why? Both `Boolean(NaN || NaN)` and `Boolean(NaN >=0 || NaN >=0)` are false.

SukkaW (changes requested):

Extensions/chrome/return-youtube-dislike.script.js
gitdev-bash: NaN translates to boolean false already

Extensions/firefox/return-youtube-dislike.script.js
```diff
@@ -71,7 +71,7 @@ function setState() {
         cLog("response from youtube:");
         cLog(JSON.stringify(response));
         try {
-          if (response.likes || response.dislikes) {
+          if (response.likes >=0  || response.dislikes >=0) {
```
gitdev-bash, line 74: NaN translates to boolean false already

gitdev-bash (changes requested):
These changes won't change anything in the function of the code.
NaN already translates to boolean false.

Anarios:
Fixed a bit differetly, but thanks for idea.

PR #182, opened 2021-12-04
fix missing styles for rate bar in userscript

NgoKimPhu:
add missing css styles for rate bar, copied over from content-style.css


Nightcaat (approved):

Anarios:
Thanks!

PR #197, opened 2021-12-05
refactor: extract hardcoded values to const

dancingpizza:
Did a little refactoring.
Extracted the hardcode state values into constants, what would be easier to develop in future.


Anarios (approved):
Thanks

Issue #201, opened 2021-12-05
Format dislikes number the same way likes are formatted


oSumAtrIX:
### Extension or Userscript?

Both

### Request or suggest a new feature!

This image by turnoffthelights.com shows how **YouTube** is formatting the dislikes count. Currently, **Return YouTube Dislike** lacks doing the same for specific locales. 
![image by turnoffthelights.com](https://www.turnoffthelights.com/support/wp-content/uploads/2019/09/taylor-swift-lover-youtube-video-color-filter-1024x610.png)

### Ways to implement this!

The front-end should format the dislikes count prior to injecting it into the HTML. 
Therefore, a function can be implemented to handle the formatting of the number.

### Can you work on this?

- [X] Yes
- [ ] No

oSumAtrIX:
Formatting fails on the user script for locale `de_DE` because the regex is wrong and should be  `/(\.0)|,0/`. 
Instead, the formatter from the user extension should be used.

PR #202, opened 2021-12-05
Format dislikes the same way likes are formatted

oSumAtrIX:
Change function `numberFormat` to match the formatting on YouTube more closely. 
In the range of millions, YouTube hides decimals, which `numberFormat` does not do yet.
Related issue: #201

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -213,32 +213,10 @@ function getVideoId() {
   return videoId;
 }
 
-function isVideoLoaded() {
-  const videoId = getVideoId();
-
-  return (
-    document.querySelector(`ytd-watch-flexy[video-id='${videoId}']`) !== null
-  );
-}
-
-function roundDown(num) {
-  if (num < 1000) return num;
-  const int = Math.floor(Math.log10(num) - 2);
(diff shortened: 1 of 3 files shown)


Extensions/UserScript/Return Youtube Dislike.user.js
```diff
@@ -221,24 +221,15 @@ function isVideoLoaded() {
...
-  const userLocales = navigator.language;
-
-  const formatter = Intl.NumberFormat(userLocales, {
-    notation: "compact",
-    minimumFractionDigits: 1,
-    maximumFractionDigits: 1,
-  });
-
-  return formatter.format(roundDown(numberState)).replace(/\.0|,0/, "");
+function numberFormat(num) {
```
UnrealValentin, line 224: maybe let the user choose what way they want it displayed?

UnrealValentin (approved):
overall good?

Extensions/UserScript/Return Youtube Dislike.user.js
oSumAtrIX: For the extension a toggle can be implemented into the pop-out, though for tampermonkey users another solution has be to created

Anarios:
Won't work for international users, locale has to be used

Issue #195, opened 2021-12-04
(Bug): YouYube Music


Ascor8522:
### Browser

Google Chrome

### Browser Version

Google Chrome 96.0.4664.45 (64 bits) (cohort: Stable)

### Extension or Userscript?

Extension

### Extension/Userscript Version

1.1.0.0

### Video link where you see the problem

https://music.youtube.com/watch?v=WEcPAs9gC70

### What happened?

While listening to music on **YouTube Music**, the extensions throws errors in the background on a loop.

```
Uncaught TypeError: Cannot read properties of null (reading 'offsetParent')
    at getButtons (return-youtube-dislike.script.js:18)
    at checkForJS_Finish (return-youtube-dislike.script.js:186)
getButtons @ return-youtube-dislike.script.js:18
checkForJS_Finish @ return-youtube-dislike.script.js:186
```

## Steps to reproduce

1. Go on YouTube Music.
2. Listen to some music
3. Check the console full of that same error

mphelp:
@Frontesque @Anarios looks like #204 closes this issue 👍

PR #204, opened 2021-12-05
Fix Issue #195: Fix bug where extension was being loaded on youtube music domains

trevorpetersen:
It seems to me that the extension should not be run on youtube music domains. 

Although the error was only happening on chrome, I also added the exclusion to Firefox for consistently.


UnrealValentin:
Is there an issue with it being loaded on yt music? Isn't it nice if it would work there too?

Nightcaat (approved):
It doesn’t seem like YouTube Music shows like/dislike ratios in the first place

Anarios (approved):

Anarios:
ty, we might look into showing dislikes in yt music later

PR #209, opened 2021-12-05
Stop calculating dislike when creator has disabled

avinashjsap:
This fix will stop updating the dislike count by checking the **_allowRatings_** property

P.S: I am not clear why we get the exact same details like counts from https://returnyoutubedislikeapi.com/votes?videoId, so I have added a delay in order to avoid race conditions when buttons are being manipulated

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -165,6 +165,12 @@ function setState() {
   fetch(`https://www.youtube.com/watch?v=${getVideoId()}`).then((response) => {
     response.text().then((text) => {
       let result = getDislikesFromYoutubeResponse(text);
+      if (result === false){
+        cLog("response from youtube:");
+        cLog("Creator has opted to hide likes and dislikes");
+        statsSet = true;
+        return;
+      }
       if (result) {
         cLog("response from youtube:");
         cLog(JSON.stringify(result));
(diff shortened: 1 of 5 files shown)


Anarios:
Nah, we'll show dislikes on videos with disabled dislikes as well.

avinashjsap:
It does not make any sense. What is the use of showing "0" all the time
image

It lacks consistency

With this update it would be like

Screenshot 2021-12-05 at 10 23 44 PM

Anarios:
It will show numbers from users of extension with next update, so it's useful only while next update is not up. Thanks for effort, and sorry

Anarios:
Race condition is there by design. Wait time would break usability for users who don't have averageRating in youtube response.

avinashjsap:
Using 100% extension likes and dislikes has the following disadvantages

First of all, artificially giving a like / dislike count based on your formula is not correct. It would give a wrong impression when a user uses this plugin and sees like and dislike count. Additional attention is needed for the user to figure out if it is an artificial count or legit count even if we visually distinguish them
Secondly, counting the number of extension views is hard
Thirdly, even if we have a clone of yt's view count algorithm, it would be a biased result based on the number of people who uses the plugin and it will have an impact on the content and its creator's revenue

Anarios:
Yes, these are known concerns, but I'm going this route. There will be a visual indication that the numbers are from extension users only.

avinashjsap:
> Yes, these are known concerns, but I'm going this route. There will be a visual indication that the numbers are from extension users only.

Are you planning to store the likes and dislikes of every user?

Anarios:
Yes

avinashjsap:
Ok, Count me in, if you want some work to be done!

Anarios:
Thanks a lot! And sorry for wasting your time with this PR - it's actually quite good. But obsolete.

avinashjsap:
Pls share your discord user info.
We could talk over there

PR #212, opened 2021-12-05
Fix dislike count increasing in Firefox.

surrealegg:
likeClicked() suppose to decrease the dislike count by 1 if the previous state is "disliked", but the function is empty on Firefox.


Anarios (approved):

Anarios:
thank you

PR #222, opened 2021-12-06
✅ Fix small Typo

Grif-Real:
Fixed a typo. Browser extension and an API that **show** you dislikes on youtube ---> Browser extension and an API that **shows** you dislikes on youtube


DARKDRAGON532:
Pretty sure `show` is correct in this context

imansour12:
show is the correct one,

Anarios:
It was already corrected the other way arounds. Some people think it should be show (me included), but most prefer shows.

PR #223, opened 2021-12-06
update rating bar on input

cupiditys:
* rating bar updates when you click like/dislike
* changed "response.likes && response.dislikes" to "response.viewCount" to prevent errors on videos with no likes
> 
> gif

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -25,6 +25,9 @@
 const LIKED_STATE = "LIKED_STATE";
 const DISLIKED_STATE = "DISLIKED_STATE";
 const NEUTRAL_STATE = "NEUTRAL_STATE";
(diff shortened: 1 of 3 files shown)


JacynoAdam:
Doesn't changing to accepting `response.viewCount` cause a problem with less accurate YouTube data overwriting API data? Specifically, I mean, as far as I could tell, YouTube reponse would always return 0 dislikes (as it's hidden everywhere now), and if the value of dislikes is not checked (`if (response.likes && response.dislikes)`) would it not cause a number with dislikes from API reponse being overwritten with data scrapped from YouTube with 0 dislikes?

Sorry if I'm missing something obvious, not very experienced with extensions.

Anarios:
Some conflicts, sorry :( Have time to resolve?

cupiditys:
there

cupiditys:
also fixed #232 as you can see

PR #234, opened 2021-12-06
Fix for Pale Moon

### LINKED ISSUE #128: (Feature Request): Pale Moon Compatibility
### Extension or Userscript?

Userscript

Is there any chance you could make the userscript compatible with Pale Moon/UXP Browsers (http://www.palemoon.org) ?

### Ways to implement this!

_No response_

### Can you work on this?

- [ ] Yes
- [ ] No

Zorgodon:
This is a minimal patch that gives an acceptable output for the [Pale Moon](http://www.palemoon.org/) browser while maintaining compatibility with Chrome and FF. PM doesn't support the compact notation needed to get the metric suffixes, but this is enough to call it 'compatible' with it. 
This will close #128 .

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -301,8 +301,6 @@ function numberFormat(numberState) {
     maximumFractionDigits: 0,
   });
 
-  cLog('This is the number format!');
-  cLog(formatter.format(roundDown(numberState)).replace(/\.0|,0/, ""));
   return formatter.format(roundDown(numberState)).replace(/\.0|,0/, "");
 }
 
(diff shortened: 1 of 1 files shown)


Anarios:
You will need to add null checks, not just remove "?."

Zorgodon:
Would something like
`return document.getElementById("menu-container") ? document.getElementById("menu-container").querySelector("#top-level-buttons-computed"): undefined ;` do? I've never coded JS before.

As for the `<!--css-build:shady-->`,  PM ignores the comment and prints likes/dislikes in black font below the ratebar.

I'm happy to close this for now and look for a solution for that too.

pav1388:
> This is a minimal patch that gives an acceptable output for the [Pale Moon](http://www.palemoon.org/) browser while maintaining compatibility with Chrome and FF. PM doesn't support the compact notation needed to get the metric suffixes, but this is enough to call it 'compatible' with it. This will close #128 .

Could adapt the script under Palemoon 29 and GreaseEmonKey 3.31.4? If so, then you can get a working script?

PR #253, opened 2021-12-09
Variable declaration and useless file

KalinkaGit:
Hello,

I changed the declaration of some variable because it's better to use 'let' instead of 'var' since ES2015.
I also removed 1 file in the Firefox extension that was empty, and also remove the import line in the manifest.json

Regards.


Anarios:
Hi, did you test it? I remember at least one place where var was used for a reason (re-inserting of script would cause an error with let)

KalinkaGit:
Hi, I tested it on Chrome and Firefox, there was just an error in the popup.js due to the Advanced Toggle that isn't implemented in the HTML, so I commented it.

I also found the error you told me about in the UserScript, indeed, and there is a problem and I fixed it in my last commit, tested it and working fine!

KalinkaGit:
Also fixed a bug due to const declaration in the future version of firefox extension, now working fine

PR #282, opened 2021-12-13
Split chrome script into several helper files

leonbubova:
## Summary
This PR introduces quite a lot of changes. Since `return-youtube-dislike.script.js` has grown quite big, I tried to split it up into several smaller, more maintainable and possibly testable files and created a file structure where I organised the new files thematically, using ES6 modules.

For most of the changes I just moved the functions into their own files and didn't change anything of the functionality itself, other than a handful of changes, but I did introduce some extra parameters to some of the functions to fix the not globally reachable parameters of some functions now.

### What
- Split up `return-youtube-dislike.script.js` into several helper files (see Screenshot for proposed file structure)
- created ES6 modules out of the helper files to make them importable and reusable throughout the project
- refactored usage of RYDTools and RYD class to be more simpler, since they didn't work that good with the new structure

### Why
Big script files can get out of hand soon and become unmaintainable and untestable. Therefore we should address this sooner rather than later.

### New File Structure
image

PR #311, opened 2021-12-15
Implement webpack to use ES6 modules

leonbubova:
## Summary
I added webpack to the project, to be able to use ES6 modules which let's us split up files into modules. This is related to #282 since the changes in there, won't work without webpack and modules.

## What
- Create package.json to handle dependencies
- Add webpack and babel to project
- Create script `npm run bundle` to enable automatic rebuilding of source files on save
- Update `CONTRIBUTE.md` with npm setup

## Why
- To be able to use ES6 modules, which increases code maintainability
- Makes it possible to (after I cleaned it up again) merge #282


leonbubova:
Is it correct that 6 approving reviewers are required for this PR? That seems a bit hefty 😅

glenlalo (commented):
It would be good if you may change the commit message to something else and not just wip

leonbubova:
> It would be good if you may change the commit message to something else and not just wip

Isn't it getting squashed automatically?

glenlalo:
@leonbubova Have you tried loading the extension in firefox? I'm getting this error: `Unsupported manifest version: 3`
In chrome works fine :)

leonbubova:
I think for Firefox you have to rename manifest.json to manifestv3.json and manifestv2.json to manifest.json. But let me double check that as well.

leonbubova:
Yes just checked, after renaming the `manifestv2.json` to `manjfest.json` it works.

glenlalo:
> I think for Firefox you have to rename manifest.json to manifestv3.json and manifestv2.json to manifest.json. But let me double check that as well.

Thanks, after renaming, works in both browsers without having to rename again 👍  Can you push the fix?

leonbubova:
> > I think for Firefox you have to rename manifest.json to manifestv3.json and manifestv2.json to manifest.json. But let me double check that as well.
> 
> Thanks, after renaming, works in both browsers without having to rename again 👍 Can you push the fix?

Just did

glenlalo (approved):

Anarios:
One of the biggest PRs to extension code so far, thanks so much.

PR #321, opened 2021-12-15
Patch: Remove unneeded function after YT-API changed and fixed loading of extension in Firefox

### LINKED ISSUE #320: (Bug): Loading RYD as a temporary extension does not work in Firefox
### Browser

Firefox

### Browser Version

latest

### Extension or Userscript?

Extension

latest

null

### What happened?

A bug happened!

glenlalo:
- Number formatting is not needed anymore, since the extension does not fetch dislike data from YT.
- Loading as temporary extension in Firefox is fixed #320


Anarios:
manifest file is for chrome. manifest v2 is used with FF, so what you did breaks chrome.

glenlalo:
@Anarios I tested in chrome and it works too.

Anarios:
@ghost, yea - but it changes manifest version, which prevents from publishing to chrome store

Anarios:
This is done with webpack now.

PR #323, opened 2021-12-15
remove unused functions from extension code

himanshudabas:
There was 1 unused function in background script `getDislikesFromYoutubeResponse()`, which was being used before 13th as a fallback to show dislikes to users.
It is not being used anywhere in the code anymore.

In the content-script there were 2 methods `isVideoNotLiked()` & `isVideoNotDisliked()` which were also not being used anymore.
content-script already have `isVideoLiked()` & `isVideoDisliked()` method which are being used now.

### DIFF
--- a/Extensions/combined/ryd.background.js
+++ b/Extensions/combined/ryd.background.js
@@ -131,37 +131,6 @@ api.storage.sync.get(null, (res) => {
 const sentIds = new Set();
 let toSend = [];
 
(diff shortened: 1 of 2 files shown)


leonbubova (approved):
Thanks for that PR. It annoyed me as well to be honest, but I didn't have the time 👍

Anarios (approved):

Anarios:
Thanks!

Issue #180, opened 2021-12-04
(Bug): Doesn't query dislikes for youtube clips


WormChickenWizard:
### Browser

Brave

### Browser Version

1.32.113 Chromium: 96.0.4664.45 (Official Build) (64-bit)

### Extension or Userscript?

Extension

### Extension/Userscript Version

1.1.0.0

### Video link where you see the problem

https://youtube.com/clip/Ugkx-vmGREh9elsBRbgrwnJqMwhQSaxr5YsW

### What happened?

Anytime that I navigate to a YouTube clip (i.e https://youtube.com/clip/*), dislikes are not queried.

mphelp:
A clip's ID looks quite dissimilar to a video ID. But fortunately, a clip's stats appear to be a carbon copy of the original video's stats. Even the hidden "averageRating":4.9558887

If you click on "watch full video" in the clip linked above, you'll be brought to https://www.youtube.com/watch?v=1Ntwv1z6WAk. 

We can archive the stats of the original video from this link instead of the clip to keep everything consistent, since the stats are identical.

Frontesque:
This is an issue with the extension only having permissions to access "youtube.com/watch?v=*" links. It isn't able to access "youtube.com/clip/*". Im not sure if the element layout is similar to how it is on the normal videos but I can look into it. This might be a simple fix.

Frontesque:
After looking into it, We'd need to find some way to get the original video id from the clip. It's not really a high priority right now as there are more important features that should be implemented first. This can be a thing for later

mphelp:
@Frontesque turns out there's always a meta tag with the exact ID

WormChickenWizard:
I'm now running version 1.1.0.2 of the extension and it still isn't showing the number of dislikes on the clips. I looked at #215 and its trying to match /clips/* when it should be /clip/* (notice there's a lack of an s).

mphelp:
@WormChickenWizard Thanks for finding my typo. Already merged in #267

PR #324, opened 2021-12-15
Fix Issue 180: clips

mphelp:
The initial load check on JS was strictly set to urls with `watch?`
Clips have urls with `/clip/...`
This PR allows the check to occur for either clips or videos and moves the URL matching from content script to manifest
re: #180

### DIFF
--- a/Extensions/combined/manifest-chrome.json
+++ b/Extensions/combined/manifest-chrome.json
@@ -19,11 +19,7 @@
   },
   "content_scripts": [
     {
-      "matches": [
-        "*://youtube.com/*",
-        "*://www.youtube.com/*",
-        "*://m.youtube.com/*"
-      ],
+      "matches": ["*://*.youtube.com/watch*", "*://*.youtube.com/clip*"],
       "exclude_matches": ["*://*.music.youtube.com/*"],
       "js": ["ryd.content-script.js"],
       "css": ["content-style.css"]
--- a/Extensions/combined/manifest-chrome.json
+++ b/Extensions/combined/manifest-chrome.json
@@ -19,7 +19,11 @@
   },
   "content_scripts": [
     {
-      "matches": ["*://*.youtube.com/watch*", "*://*.youtube.com/clip*"],
+      "matches": [
+        "*://youtube.com/*",
+        "*://www.youtube.com/*",
+        "*://m.youtube.com/*"
+      ],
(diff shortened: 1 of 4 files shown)


DARKDRAGON532 (approved):

mphelp:
@Anarios we could probably simplify the content script matches down to 
`        "*://youtube.com/watch?*",

        "*://www.youtube.com/watch?*",`
since `"*://*.music.youtube.com/*"` is already excluded

Anarios:
@mphelp, I'm not entirely sure. You tested it? If yes - I think this could be merged

mattphelps-8451:
Just tested both combined and Userscript on both videos and clips. Everything went smoothly.

DARKDRAGON532 (approved):
resolve merge conflicts btw

Anarios:
I could resolve conflicts in my branch, but then you'd lose contributor stats (if you care about them). So whichever way is more preferable for you. Thanks!

UnderscoreX5 (approved):

Frontesque:
Please resolve conflicts before being merged

mattphelps-8451:
@Anarios @Frontesque @UnderscoreX5 Resolved conflicts!

mphelp:
@Anarios @Frontesque @UnderscoreX5 Resolved conflicts!

Extensions/combined/manifest-chrome.json
```diff
@@ -19,11 +19,7 @@
   },
   "content_scripts": [
     {
-      "matches": [
-        "*://youtube.com/*",
-        "*://www.youtube.com/*",
-        "*://m.youtube.com/*"
-      ],
+      "matches": ["*://*.youtube.com/watch*", "*://*.youtube.com/clip*"],
```
Anarios, line 22: Last time we changed permissions - we lost over 100k users in one day :( Not sure if we should go with it. I will investigate if may be this can be moved to optional permissions

Anarios: So that a dynamic prompt appears when /clip is visited

Anarios: On the other hand - if old permissions are *://youtube.com/* - we should just keep it.

mphelp:
@Anarios I kept the existing manifest url matching as it was before like you requested

Anarios:
Thanks!

PR #333, opened 2021-12-17
Update  build scripts, .gitignore and docs

leonbubova:
## What
- Rename `npm run bundle` to `npm run build` and remove file watcher from config
- Add `npm start` script with file watcher in options
- Update docs with new/renamed scripts
- Add build file to .gitignore

## Why
- Prevent build files from accidentally being committed
- Make build and start scripts follow more common naming conventions to make it easier for new contributors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ yarn-error.log*
 *.sw*
 
 node_modules
+
+# Build files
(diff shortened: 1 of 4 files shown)


Anarios (approved):
Thanks!

PR #348, opened 2021-12-19
Add a security FAQ

CallMeAlexO:
I've noticed that as the extension gains popularity, more and more people are pouring over with legitimate security concerns - and they open the same issues asking the same questions.. So I figured it'd be nice to have a security FAQ.


leonbubova (approved):
Nice writeup 👌🏻

SECURITY.md
```diff
@@ -1,30 +1,30 @@
...
 
-## What information do you have, exactly?
+### What information do you have, exactly?
 
 Just the video ID. Not your comments, not your username, not who you've shared the video with, not any additional metadata. Nothing. Just the video ID. 
 
-## How is my IP stored?
+### How is my IP stored?
 
 The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. This is done to prevent database vandalism.
```
UnderscoreX5, line 17: on a hard or solid-state drive perhaps?

SECURITY.md
```diff
@@ -1,30 +1,30 @@
...
 
-## How is my IP stored?
+### How is my IP stored?
 
 The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. This is done to prevent database vandalism. 
 
-## I heard some discussion over OAuth, and access to my YouTube account!
+### I heard some discussion over OAuth, and access to my YouTube account!
 
 This feature is optional, and very much opt-in. If you are a YouTube creator, and would like to share your dislike stats with us, you can. The way OAuth was structured, is it's actually very secure. You can revoke access to your account at any time, and can give very specific permissions to us. We will not ask for any permissions that aren't required. We'll only ask for permissions to view your video stats.
```
UnderscoreX5, line 21: Could add a link to https://en.wikipedia.org/wiki/OAuth so that anyone that's curious enough can double check on their own

FAQ.md
```diff
@@ -16,3 +16,6 @@ The extension collects the video id of the video you are watching, fetches the d
 
 ### **4. What will happen after the YouTube API stops returning the dislike count?**
 The backend will switch to using a combination of archived dislike stats, estimates extrapolated from extension user data and estimates based on view/like ratios for videos whose dislikes weren't archived and for outdated dislike archives.
+
+## I have security concerns
```
UnderscoreX5, line 20: I have concerns regarding the security and anonimity of this extension (could perhaps sound better)

UnderscoreX5 (commented):
Other than the 3 small changes I would personally make, I think this is a solid security FAQ that should be implemented.

SECURITY.md
```diff
@@ -0,0 +1,30 @@
+# Security
+
+### Are you tracking my viewing history?
+
+No. The extension's code is public and you can see it for yourself. The only information being sent is the video ID. There are no additional headers being sent. Over the communication layer, your public IP will be exposed to the server, as well as the time when the request was made. However, none of these are uniquely identifying you in any way. Assuming a zero-trust environment, the best we could get is a dynamic IP. Which, today is yours, tomorrow is your neighbor's. If you're really worried about your IP being traced, you probably already use a VPN.
```
himanshudabas, line 5: > The only information being sent is the video ID.

can we instead say `The only information being sent is the video ID, which is required to fetch the dislike count for the videos.`, so that it doesn't give an impression that the video ID being sent is for tracking.

SECURITY.md
```diff
@@ -0,0 +1,30 @@
...
+
+No. When you dislike a video, we create a unique (but random and non-identifiable) ID for you. This is done to prevent botting. But again, it cannot uniquely identify you.
+
+### What information do you have, exactly?
+
+Just the video ID. Not your comments, not your username, not who you've shared the video with, not any additional metadata. Nothing. Just the video ID. 
+
+### How is my IP stored?
+
+The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. This is done to prevent database vandalism.
```
himanshudabas, line 17: I don't think we even keep IPs in plain text. I think we keep hashes of IPs.
@Anarios could you please clarify?

SECURITY.md
```diff
@@ -0,0 +1,30 @@
...
+
+Just the video ID. Not your comments, not your username, not who you've shared the video with, not any additional metadata. Nothing. Just the video ID. 
+
+### How is my IP stored?
+
+The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. This is done to prevent database vandalism. 
+
+### I heard some discussion over OAuth, and access to my YouTube account!
+
+This feature is optional, and very much opt-in. If you are a YouTube creator, and would like to share your dislike stats with us, you can. The way OAuth was structured, is it's actually very secure. You can revoke access to your account at any time, and can give very specific permissions to us. We will not ask for any permissions that aren't required. We'll only ask for permissions to view your video stats.
```
himanshudabas: I think it's worth mentioning that OAuth hasn't been implemented yet.

SECURITY.md
```diff
@@ -0,0 +1,30 @@
...
+
+The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. This is done to prevent database vandalism. 
+
+### I heard some discussion over OAuth, and access to my YouTube account!
+
+This feature is optional, and very much opt-in. If you are a YouTube creator, and would like to share your dislike stats with us, you can. The way OAuth was structured, is it's actually very secure. You can revoke access to your account at any time, and can give very specific permissions to us. We will not ask for any permissions that aren't required. We'll only ask for permissions to view your video stats. 
+
+### How can I trust this dislike count?
+
+Short answer is: you can't. We've implemented anti-botting measures, but at the end of the day, it's up to you to decide whether you trust our count or not.
```
himanshudabas, line 25: Maybe be a little less direct here :)
instead write something like
```We have implemented measures to prevent bot attacks, and we are gonna continue to work on improving the effectiveness of the bot prevention system. This will help us keep the dislike count as a good representative of the actual count. Of course, it can never be 100% accurate. So it's up to you to decide whether you trust the count or not.```

File: FAQ.md (hunk 3 above)
Comment (CallMeAlexO) [reply FAQ.md]: Improved phrasing

File: SECURITY.md (hunk 2 above)
Comment (CallMeAlexO) [reply SECURITY.md]: Great idea! Added.

File: SECURITY.md (hunk 1 above)
Comment (CallMeAlexO) [reply SECURITY.md:17]: I think it's pretty clear. In this case hard drive is a synonym for any storage device.

File: SECURITY.md (hunk 4 above)
Comment (UnderscoreX5) [reply SECURITY.md:5]: I agree

File: SECURITY.md (hunk 6 above)
Comment (UnderscoreX5) [reply SECURITY.md]: Yeah, maybe change thefrasing like "This feature _will be_ optional" and just clarify that it's not out yet.

Comment (CallMeAlexO) [reply SECURITY.md]: The goal of the security FAQ is to cover the security aspect of things, it isn't the changelog. Hopefully it shouldn't be updated that often. I think adding this in ahead of time can save a lot of people a lot of headaches in advance :)

File: SECURITY.md (hunk 4 above)
Comment (CallMeAlexO) [reply SECURITY.md:5]: Good catch!

File: SECURITY.md (hunk 7 above)
Comment (CallMeAlexO) [reply SECURITY.md:25]: I like this phrasing more, added!

Comment (UnderscoreX5) [reply SECURITY.md]: We have implemented measures to prevent bot attacks and are gonna continue to work on improving the effectiveness of the bot prevention system: this will help us keep the dislike count as a good representative of the actual count. Of course it will never be 100% accurate so it's up to you to decide whether you trust the count or not.

File: SECURITY.md [hunk 8]
```diff
@@ -18,7 +18,7 @@ The backend keeps IP addresses in volatile memory (RAM) only. These addresses ar
 
 ### I heard some discussion over OAuth, and access to my YouTube account!
 
-This feature is optional, and very much opt-in. If you are a YouTube creator, and would like to share your dislike stats with us, you can. The way OAuth was structured, is it's actually very secure. You can revoke access to your account at any time, and can give very specific permissions to us. We will not ask for any permissions that aren't required. We'll only ask for permissions to view your video stats. 
+This feature is optional, and very much opt-in. If you are a YouTube creator, and would like to share your dislike stats with us, you can. The way [OAuth](https://en.wikipedia.org/wiki/OAuth#:~:text=but%20without%20giving%20them%20the%20passwords.) was structured, is it's actually very secure. You can revoke access to your account at any time, and can give very specific permissions to us. We will not ask for any permissions that aren't required. We'll only ask for permissions to view your video stats.
```
UnderscoreX5, line 21: not sure whenever it's intentional or not but the link looks really messy and just lands on the same page as https://en.wikipedia.org/wiki/OAuth

SECURITY.md
UnderscoreX5: yeah but it could confuse people that want to partecipate into the OAuth thing + it takes no time to change once OAuth is rolled out

SECURITY.md
CallMeAlexO: It's intentional :) 
[Text Fragments](https://wicg.github.io/scroll-to-text-fragment/) are a new W3C proposal. Currently supported on most new browsers. It allows us to highlight text on the page. I highlighted the part where it says "without giving them the passwords". If you look at the page itself, that text should be highlighted in purple.

SECURITY.md
himanshudabas: @CallMeAlexO could you please update it again with the phrasing UnderscoreX5 has suggested.

SECURITY.md
UnderscoreX5: It doesn't show here on Firefox but since I use a hardened version of it I suppose it's just my issue so no worries then

SECURITY.md
CallMeAlexO: Done.

SECURITY.md
CallMeAlexO: Done.

UnderscoreX5 (approved):
Unless there are some pretty small issues we didn't catch I think this is good to go! Well done!

SECURITY.md
Anarios: Yes, we do cap IP hashes. Un-hashed IPs are stored in-memory only, so this should be updated

Anarios:
We store IP hashes, so this needs to be mentioned as well.

SECURITY.md
himanshudabas: @CallMeAlexO please mention that we store only `IP Hashes` to prevent data vandalism from bad actors.
maybe add a wiki link as well for hashing.

SECURITY.md
himanshudabas: marking this as resolved as it's no longer valid. refer https://github.com/Anarios/return-youtube-dislike/pull/348/files#r771971987

himanshudabas (approved):
looks good to me.

DARKDRAGON532 (changes requested):
`SECURITY.md` is usually for security politcy, rename the file to `SECURITY-FAQ.md` or something similar

DARKDRAGON532 (approved):

Anarios:
Thanks!

Anarios:
I removed the "hases can not be reversed" part, because I wouldn't be so sure when dealing with IPs (rainbow tables, etc). So, it complicates things a bit, but doesn't make it undoable. 

And I doubt there is anything we can do. Always open to ideas, though.

SECURITY-FAQ.md
```diff
@@ -0,0 +1,30 @@
+# Security
+
+### Are you tracking my viewing history?
+
+No. The extension's code is public and you can see it for yourself. The only information being sent is the video ID, which is required to fetch the dislike count for the videos. There are no additional headers being sent. Over the communication layer, your public IP will be exposed to the server, as well as the time when the request was made. However, none of these are uniquely identifying you in any way. Assuming a zero-trust environment, the best we could get is a dynamic IP. Which, today is yours, tomorrow is your neighbor's. If you're really worried about your IP being traced, you probably already use a VPN.
+
+### Can you uniquely identify me if I dislike?
+
+No. When you dislike a video, we create a unique (but random and non-identifiable) ID for you. This is done to prevent botting. But again, it cannot uniquely identify you.
```
ajayyy, line 9: How is a unique id "non-identifiable"?

```suggestion

Yes. When you dislike a video, we create a randomly generated unique ID for you that is not tied to your Google account. This is done to prevent botting.
```

SECURITY-FAQ.md
```diff
@@ -0,0 +1,30 @@
...
+
+### What information do you have, exactly?
+
+Just the video ID. Not your comments, not your username, not who you've shared the video with, not any additional metadata. Nothing. Just the video ID. 
+
+### How is my IP stored?
+
+The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. We hash the IP addresses, and that's stored instead. This is done to prevent database vandalism. A hash is a one-way function, so your IP address cannot be guessed from the hash.
```
ajayyy, line 17:
```suggestion
The backend keeps IP addresses in volatile memory (RAM) only. These addresses aren't stored on a hard drive, and therefore aren't logged. We hash the IP addresses, and that's stored instead. This is done to prevent database vandalism. 

A hash is a one-way function, requiring guess and check to determine the original data. Since there are a limited set of IPV4 addresses (2^32), it is possible to reverse this hash, but hashing increases the effort required to do so.
```

ajayyy:
Oh, this is merged. Will make PR

CagesThrottleUs:
> And I doubt there is anything we can do. Always open to ideas, though.

Okay, I am no security expert and what I am going to say is from just normal searching "how to prevent reversing of a hash". 

So to prevent reverse hashing, I thought of treating it as a problem of password protection, where you would be protecting the dynamic IP with same level as a password. A standard hash function like SHA-256 etc. can be used but a lot of suggestion has been against it. 

Something that caught my eye is Argon2. It has been recommended via [PHC](https://www.password-hashing.net). But to implement Argon2, there are more security stuff like Salt, Parallelism, etc. A lot of this can be seen from [this webpage](https://argon2.online). I recommend that someone actually look into specification, time-cost etc. I will help with the limit of my abilities if required.

Anarios:
@CagesThrottleUs difference for IPs is that you know possible value range. So methods described here do not work so well for IPs. It's better with IPV6, but for IPV4 value range is quite limited.

CagesThrottleUs:
> @CagesThrottleUs difference for IPs is that you know possible value range. So methods described here do not work so well for IPs. It's better with IPV6, but for IPV4 value range is quite limited.

I see, perhaps the idea of salts may work with the current existing stuff. Anyway, hope you find something useful and thank you for taking time to view this.

himanshudabas:
> I see, perhaps the idea of salts may work with the current existing stuff. Anyway, hope you find something useful and thank you for taking time to view this.

Even if we add salts to the hashed IPs, that salt has to be consistent for all the IPs, else we won't have a way to map new requests to our hashed + salted IPs in our db.
We could use some function which takes the IP itself to compute a unique salt for each ip address.
that would certainly make the job difficult for hackers to map the hashed + salted addresses to their corresponding IPv4 addresses.

If backed gets somehow compromised, hackers could still get access to the function itself.
IPv4 are just over 4 billion. It'd be a relatively easy task for them to compute hash + salt of all available IPv4 addresses.

But I get your point, It's certainly better than simply storing sha512 hashes, as they are pretty easy & fast to compute.
`bcrypt` is better than `sha512`, `Argon2` is even better than `bcrypt` for this use case.

It is worth looking into to strengthen the backend.

UnderscoreX5:
I do believe that hashing the IPs is surely beneficial but at the same time it's kinda a useless measure if you get what I mean. IPs aren't really ideal for hashing since they aren't infinite and there is actually a limited bunker of them. Furthermore what you are stealing is an IP connected to a random ID at max, there's little reason to since we already have a pool of more than a million users and that keeps growing. We can surely do it because more safety is better but stealing a random IP address of someone you can't even identify is like DDOSSING a random IP address you made up following the real parameters... useless.

Issue #350, opened 2021-12-19
(Feature Request): Version Number


UnderscoreX5:
### Extension or Userscript?

Extension

### Request or suggest a new feature!

Adding the version number to the extension could be really useful for troubleshooting, changelogs and safety + it's not hard to do.

### Ways to implement this!

Perhaps it could be formatted as v0.1 where v1 is the stable relase, but any other method works. It could be added in the extension's menu in the bottom right corner.

### Can you work on this?

- [x] Yes
- [ ] No

himanshudabas:
Perhaps something like this:

![image](https://user-images.githubusercontent.com/20219940/146681578-0c787cd9-f85d-4a00-8415-ed464560d2bd.png)


are you working on this?

UnderscoreX5:
Yes. I set up a pretty raw version of it. I tried making it show up above the buttons since it also looked good but it might be too distracting.

cyrildtm:
Implemented in commit db666735e4edb4af69b51f352df25013e5119fb6

Anarios:
Was implemented

PR #351, opened 2021-12-19
Add version number to extension's menu

UnderscoreX5:
Dinamically takes the manifest's version number and displays it under the credits text. **Full compatibility with Firefox and Safari is the current priority.** Placement of the number must be changed to be in the bottom right corner. Currently broken i guess.

### DIFF
--- a/Extensions/combined/popup.html
+++ b/Extensions/combined/popup.html
@@ -13,6 +13,7 @@
       <svg width="48" viewBox="0 0 24 24"><path d="M14.9 3H6c-.9 0-1.6.5-1.9 1.2l-3 7c-.1.3-.1.5-.1.7v2c0 1.1.9 2 2 2h6.3l-.9 4.5c-.1.5 0 1 .4 1.4l1.1 1.1 6.5-6.6c.4-.4.6-.9.6-1.4V5c-.1-1.1-1-2-2.1-2zm7.4 12.8h-2.9c-.4 0-.7-.3-.7-.7V3.9c0-.4.3-.7.7-.7h2.9c.4 0 .7.3.7.7V15c0 .4-.3.8-.7.8z" fill="red"/><path d="m8 12.5 5.1-2.9L8 6.7v5.8z" fill="#fff"/></svg>
       <h1>Return YouTube Dislike</h1>
       <p>by Dmitrii Selivanov & Community</p>
+      <p>v2.0.0.3</p>
 
       <button id="link_website">Website</button>
       <button id="link_github">GitHub</button>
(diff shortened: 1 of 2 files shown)


Extensions/combined/popup.html
```diff
@@ -13,6 +13,7 @@
       <svg width="48" viewBox="0 0 24 24"><path d="M14.9 3H6c-.9 0-1.6.5-1.9 1.2l-3 7c-.1.3-.1.5-.1.7v2c0 1.1.9 2 2 2h6.3l-.9 4.5c-.1.5 0 1 .4 1.4l1.1 1.1 6.5-6.6c.4-.4.6-.9.6-1.4V5c-.1-1.1-1-2-2.1-2zm7.4 12.8h-2.9c-.4 0-.7-.3-.7-.7V3.9c0-.4.3-.7.7-.7h2.9c.4 0 .7.3.7.7V15c0 .4-.3.8-.7.8z" fill="red"/><path d="m8 12.5 5.1-2.9L8 6.7v5.8z" fill="#fff"/></svg>
       <h1>Return YouTube Dislike</h1>
       <p>by Dmitrii Selivanov & Community</p>
+      <p>v2.0.0.3</p>
```
himanshudabas, line 16: instead of hardcoding, you might wanna use `chrome.runtime.getManifest().version;`

himanshudabas (changes requested):

himanshudabas:
cc https://github.com/Anarios/return-youtube-dislike/issues/350

Extensions/combined/popup.html
Anarios: and not chrome, but cross-browser version

UnderscoreX5: Yeah that's better. Will change it as soon as I can.

UnderscoreX5: I know that the namespace for firefox and safari is 'browser' but is there a cross-browser one fir both chromium and firefox/safari?

himanshudabas: this is how we do this in bg-script:

https://github.com/Anarios/return-youtube-dislike/blob/main/Extensions/combined/ryd.background.js#L11

try to test it in both browsers.

UnderscoreX5: It shouldn't work or at least not as efficiently, perhaps implementing it this way is better, coming up in seconds with a possible fix.

UnderscoreX5: @himanshudabas can you propose the 'chrome.runtime.getManifest().version;' change you did so I can double check on my firefox as well?

himanshudabas: popup.html

```html
<p id="ext-version"></p>
```

popup.js

```javascript
function initConfig() {
  initializeVersionNumber();
  ...
}

...

function initializeVersionNumber() {
  const version = chrome.runtime.getManifest().version;
  document.getElementById('ext-version').innerHTML = 'v' + version;
}
```

Anarios:
> @Anarios I just tested chrome.runtime.getManifest().version; in firefox. I don't know why but it works (:

I think it works in later FF versions, but might break in older ones, which would otherwise work. I'm not sure though.

UnderscoreX5:
I suppose that the best way is then to use both browser and chrome namespaces just to be sure it works on all browsers (it works on firefox but what about safari?). Will try something tomorrow after work.

Anarios:
I think what we do in background.js can work here as well

UnderscoreX5:
Yeah it probably works just gotta translate it to html tho

Extensions/combined/popup.html
UnderscoreX5: I implemented this for now.

UnderscoreX5:
The current version doesn't work on my firefox but I think I've misplaced the function.

Anarios (approved):

PR #356, opened 2021-12-20
Clarify identifiable information

ajayyy:
A unique id is identifiable, it just isn't tied to other accounts.


UnderscoreX5 (approved):
Good catch.

Issue #328, opened 2021-12-16
(Feature Request): Documentation for integrations


bbilly1:
### Extension or Userscript?

Both

### Request or suggest a new feature!

I've been working on [TubeArchivist](https://github.com/bbilly1/tubearchivist/), a selfhosted YouTube media server and downloader. So I was wondering what your stance is on other services integrating in this API, both for getting ratings as well as posting back to your API. 

I know, your API is open and it's easy enough to reconstruct the requests, but I still wanted to ask.

### Ways to implement this!

So this will require a minimal amount of documentation to get started, for example:
- First a mention that this is *ok* to do for responsible netizens.
- API endpoints, expected return data and how the queries should be formed.
- Expected error messages
- Any restrictions like rate limiting
- Requirements for your monitoring like user agent string or headers

I know, early stages and things might change, but I think as long as the changes are documented and announced with some time, this would be ok.

### Can you work on this?

- [X] Yes
- [ ] No

leonbubova:
Disclaimer: I am not sure what @Anarios stance is on "Is it ok to use the API for third-party apps/outside use" but if it's allowed/encouraged I would love to see an API documentation publicly available 👍🏻

Anarios:
http://returnyoutubedislikeapi.com/swagger

Not documentation, but it's something.

Third party use is encouraged. There are a couple of requirements. 1. It should be clearly attributed. 2. There are rate limits of 100 per minute and 10 000 per day (but it can be adjusted if you contact me)

bbilly1:
Thanks for your response. That sounds very reasonable, and yes, rate limiting shouldn't be a problem. I can make a pullrequest later against the readme to add a "API Documentation" section with some *getting started* info, to make sure you don't get more of the same questions here.

Anarios:
I'd appreciate that.

bbilly1:
There is another merge request describing a POST request and a python example link. Closing this issue now.

PR #357, opened 2021-12-20
add API documentation with get vote example

bbilly1:
This is a follow up for #328, basic API documentation starting with some general info and an example to `get` votes. 

To `post` votes, I think I understand what you are trying to do with your puzzle solver, but I doubt that this is feasible to integrate into third party applications like that. Maybe in the future it could be worth it to think about an JWT authentication system or similar so people could make properly authenticated requests.

I'm assuming you have your cloudflare rate limiter set to send standard status code `429`? I didn't want to DOS you just to find out. :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,37 @@ and estimates based on view\like ratios.
 
 You can learn more at our website at: [returnyoutubedislike.com](https://www.returnyoutubedislike.com/)
 
+## API documentation
+
+Third party use of this open API is allowed with the following restrictions:
+
+- **Attribution**: This project should be clearly attributed with either a link to this repo or a link to [returnyoutubedislike.com](https://returnyoutubedislike.com/).
+- **Rate Limiting**: There are per client rate limits in place of 100 per minute and 10'000 per day. This will return a *429* status code indicating that your application should back off.
+
+The API is accessible over the following base URL:  
+https://returnyoutubedislikeapi.com  
+
+List of available endpoints is available here:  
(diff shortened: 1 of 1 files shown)


himanshudabas:
Purpose of POW puzzle is to deter people from casting sever votes. They can still cast multiple votes with a good enough hardware, but there are other measures in place, like IP Hashing to determine if someone is trying to cast a duplicate vote.
<br>
> it could be worth it to think about an JWT authentication system

Yes, you are right. we discussed this as well, but Anarios didn't want to force registration on people at this early stage. I think that is understandable because most of the people think that RYD extension is doing some sort of magic to fetch the dislikes and registration would only steer the users away.

In near future, Yes. JWT tokens would be much better.
<br>
> but I doubt that this is feasible to integrate into third party applications like that

Could you please explain what you mean by this?
my understanding is, if someone is making requests from client side, they can implement the puzzle solver at client side.

Here's the code of [puzzle solver](https://github.com/Anarios/return-youtube-dislike/blob/main/Extensions/combined/ryd.background.js#L184).

bbilly1:
Yes, I get the point of the puzzle, and I also get not to enforce authentication to encourage faster adoption, and as mentioned, that can still be implemented in the future. 

So yes, the puzzle can be implemented client side, just something that will deter developers if they have to do it themself. :-)

Maybe I can come up with a Python example later so others won't have to.

Anarios (approved):

PR #359, opened 2021-12-21
added a UI for a user to see if the api is online or offline

thereaper90917:
as you can see a user can now check if the api is online for them this should help with troubleshooting going forward(as some have issues some don't) also the function only executes if the user opens the extension


UnderscoreX5:
Nice. I guess I am gonna move the version number to the bottom right corner then since it would be too messy to have it near your text

oSumAtrIX:
I am not sure if the server image is license-free. We should consider, using [Font Awesome](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=server) for the server and other icons. This will also help with different scalings.

thereaper90917:
> Nice. I guess I am gonna move the version number to the bottom right corner then since it would be too messy to have it near your text

that would be cool!

thereaper90917:
> I am not sure if the server image is license-free. We should consider, using [Font Awesome](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=server) for the server and other icons. This will also help with different scalings.
def something to consider more so the scaling also gave me an idea actually

thereaper90917:
as @oSumAtrIX suggested i removed the png and replaced it with font awesome minimal(not to clutter the entire code base) also added as you can see from screenshots that the server shows the same color to indicate online or offline (thanks for the idea @oSumAtrIX

oSumAtrIX:
Probably using grey might look better than red, which is already kind of reserved for the dislike button icon. Though, this is just my opinion.
image

thereaper90917:
> Probably using grey might look better than red, which is already kind of reserved for the dislike button icon. Though, this is just my opinion. image

hmm that does look good though i want to point attention to it as it could resolve issues quicker because it grabs their attention. as it seems important but i do like the color though the grey

UnderscoreX5:
I like the red and green ones more since they grab your attention immediately. We shouldn't worry about the red being too much since hopefully the service is almost always online.

Anarios:
Do we need entire font awesome just for one icon? 

And I also see lots of unrelated changes - do they have to be in this PR?

oSumAtrIX:
> Do we need entire font awesome just for one icon?

At some point probably the extension will make use of more characters at some point. For now the font can be minified/stripped to what is necessary, if filesize matters.

thereaper90917:
> Do we need entire font awesome just for one icon?
> 
> And I also see lots of unrelated changes - do they have to be in this PR?

to be honest most of those changes are from the npm build command. the only true line of code is in popup.js from line 85 to 101 and i believe just to elements for the UI. the font awesome i did exclude a few fonts to make it as lightweight as possible to get it to function but as @oSumAtrIX stated this could be reused in the future for more icon usage maybe? i could try and make it more minified.

Anarios:
Oh, I didn't see that it's in /dist folder, my bad.

Anarios:
Could you please pull latest from main - it has /dist and other generated files added to gitignore.

thereaper90917:
> Could you please pull latest from main - it has /dist and other generated files added to gitignore.

done removed fontawesome using the server,svg from fontawesome with filter for color change.

Anarios (approved):

PR #404, opened 2022-01-02
Fix #399

### LINKED ISSUE #399: (Bug): Privacy concern/question
### Browser

Firefox

### Browser Version

latest

### Extension or Userscript?

Extension

latest

https://www.youtube.com/watch?v=GpMP6Nh3FvU

### What happened?

I noticed that the extension issues a POST to the backend with an array of video ids, taken from the suggested videos that youtube generates for every video we click.

Why does the extension do this?

Those videos suggestions are very private data, as they could very easily be used to create strong viewing profiles, assuming the backend could be able to identify/cohort users - which is perhaps not impossible with some fingerprinting.

I also read the [FAQs](https://github.com/Anarios/return-youtube-dislike/blob/main/SECURITY-FAQ.md#are-you-tracking-my-viewing-history) and could not find any mentioning of this behaviour? It seems to imply only the ID of the video's we watch are sent.

Code: https://github.com/Anarios/return-youtube-dislike/blob/7fd4904c3924ed9cd002b7aaa5f0b642174ee553/Extensions/combined/ryd.background.js#L42

DARKDRAGON532:
resolve #399

### DIFF
--- a/Extensions/combined/ryd.background.js
+++ b/Extensions/combined/ryd.background.js
@@ -38,21 +38,6 @@ api.runtime.onMessage.addListener((request, sender, sendResponse) => {
       })
       .catch();
     return true;
-  } else if (request.message == "send_links") {
-    toSend = toSend.concat(request.videoIds.filter((x) => !sentIds.has(x)));
-    if (toSend.length >= 20) {
-      fetch(`${apiUrl}/votes`, {
-        method: "POST",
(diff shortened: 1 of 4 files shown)


Anarios:
Not merging this for now since backgroud code will be moved to content script and this code will be reused for displaying rates for each thumbnail

PR #410, opened 2022-01-08
Extension popup layout

PickleNik:
Made extension popup more compact.
Current version on Github doesn't seem to work, I'll wait for a fix and rebase as needed.
*before*
https://user-images.githubusercontent.com/31113245/148651284-ccf026d6-20a3-4adc-863e-c0fe3a539a23.mov
*after*
https://user-images.githubusercontent.com/31113245/148651292-2c55074d-a285-4122-9ba5-72a59b3b7749.mov


GeneticallyModifiedAlex:
I think the status being in the top right would be better. Just the word online that has a hover selector that trigers the displaying of the labeling clarifying that it is the API status. 
what do you think?

PickleNik:
> I think the status being in the top right would be better. Just the word online that has a hover selector that trigers the displaying of the labeling clarifying that it is the API status. what do you think?

Do you mean top-left? Settings view toggle is in the top right. (see the *after* video)
I think top-left or somewhere on the bottom next to extension version would be good, and use the space for a `Troubleshooting` link button.

GeneticallyModifiedAlex:
> I think the status being in the top right would be better. Just the word online that has a hover selector that trigers the displaying of the labeling clarifying that it is the API status. what do you think?



> the top right. (see the _a_

i do . sorry is was editing my original comment instead of making a new one

Anarios:
Amazing work, thanks

Issue #388, opened 2021-12-30
(Feature Request): Inform users with problems to restart the plugin


cyrildtm:
### Extension or Userscript?

Extension

### Request or suggest a new feature!

In the popup menu, below the credits, add a short message that tells the user if they have problem seeing the numbers, try disabling and re-enabling the plugin before going to the extension store and leaving a bad review. 

The [two paragraphs in the README](https://github.com/Anarios/return-youtube-dislike/issues/186) are insightful, but may be too long for general users, although a link maybe helpful further explaining. 

Let's make those 0.1% users more satisfied eh?

![1](https://user-images.githubusercontent.com/5927301/147786403-e1744271-cee6-4cd1-a717-7b501fb03884.PNG)
![2](https://user-images.githubusercontent.com/5927301/147786404-64055ac9-942a-4ac2-9156-dbbacae617d9.PNG)
![3](https://user-images.githubusercontent.com/5927301/147786406-1ba11372-f63e-43db-b5b4-ad1a7728c160.PNG)

### Ways to implement this!

```
PROBLEMS? 
If you have stopped seeing dislike count numbers, try disabling and re-enabling the plugin. See more details here: <link>
```

### Can you work on this?

- [ ] Yes
- [ ] No

PR #419, opened 2022-01-09
Added tips(resolves issue #388)

Cequallium:
This will add customizable tips that randomly render every time the popup is opened
There is one tip currently: "If you have stopped seeing the dislike count, try disabling and re-enabling the plugin."

The issue this resolves: https://github.com/Anarios/return-youtube-dislike/issues/388

image

### DIFF
--- a/Extensions/combined/popup.css
+++ b/Extensions/combined/popup.css
@@ -49,8 +49,8 @@ button:hover {
   bottom: 1.15rem;
   right: 1.15rem;
   background: var(--secondary);
-  border-radius: .25rem;
-  padding: .25rem .5rem;
+  border-radius: 0.25rem;
+  padding: 0.25rem 0.5rem;
 }
 
 #advancedToggle {
@@ -58,7 +58,8 @@ button:hover {
   background: none;
   box-shadow: none;
   color: var(--lightGrey);
-  top: 26px; right: 26px;
+  top: 26px;
+  right: 26px;
   padding: 2px;
   z-index: 69;
   height: 2rem;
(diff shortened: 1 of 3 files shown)


sy-b:
I would rather prefer a link to troubleshooting guide there.

Cequallium:
awesome! that would simplify it a lot

PickleNik (changes requested):
nevermind

PickleNik (approved):

cyrildtm:
> I would rather prefer a link to troubleshooting guide there.

I only partially agree. Let's acknowledge the fact that "most" people wouldn't want to spend any fraction of a second to troubleshoot. If they had to, they tend to be lazy and NOT follow any links. That's about a normal person who's not really interested into spending extra time reading what's going on under the hood. 

Most recent example: #434  

If I can't see a direct, concise message that covers 99% of the reason and works 99% of all time when I run into a trouble, I'm gonna go to "Chrome Extension Store" and find "Support" there, or leave a bad review. 

On the other hand, a "Help" button may help those who are curious and who would bother.

sy-b:
>  "most" people wouldn't want to spend any fraction of a second to troubleshoot. If they had to, they tend to be lazy and NOT follow any links.

We had people on discord asking what they had to do after downloading the extension, they were probably confused because there was near to nothing when they (at that time) clicked on the extension icon. So, I guess - people at least click on the extension icon. Not every extension works completely in the background.

If they ever encounter any problem, most will at least bother to click the icon. That should be enough. For that, instead of having just a "Help" button, we can have a diagnosis report embedded in pop.html. Something like the one that was planned for the RYD website.

For now the help button seems fine.

---
> ... and here is my question: Which paragraph from above did you really read?

all

cyrildtm:
@sy-b That's exactly my point. If someone is so confused that they can't figure out how this plug-in works out of the box without configuration, then how can we trust them to see, recognize and click on the "Help" button and sift through the long list of instructions? Big companies like M*******t and D**l have very large support forums but often the solutions are just too buried and too convoluted that ordinary users simply give up trying to find the correct answer, but not before making some very bad comments and becoming frustrated and depressed. 

Don't forget you are the 0.1% of the whole population who chose to read the whole thing. I hid my extra lengthy and convoluted comment for a purpose. Again, not everyone who wants to use this plugin wants to spend extra time knowing how it works. 

Just make your "help message" as easy to read and as standing out as possible. This extension has its potential to reach wide public awareness and some good sales technique can definitely help.

Anarios (approved):

Anarios:
Thanks

PR #432, opened 2022-01-10
use webpack for 4 entry scripts

UglyDonkey:
Changed to use webpack on all 4 entry scripts to allow for better code organization in those too


Anarios (approved):
Thanks

Issue #387, opened 2021-12-30
(Feature Request): Plugin options


cyrildtm:
### Extension or Userscript?

Both

### Request or suggest a new feature!

Implement a user option framework so one can change their preferences. Possible use cases are here (I'm counting open issues only): 

Percentage- 
#211 Display percentages by default (I think the point is to work out the math for the user without waiting for mouse hover)
Colors- 
#144 #186 Display a colored percentage besides video title
#218 Modify the ratio bar color scheme and include old-style green-red (maybe custom color code as well)
Number format- 
#201 #364 #407 Format the button counter numbers the way you like: compact & approximated or full exact; also cover more locales

### Ways to implement this!

Storage framework is implemented in #410. Tracking tasks: 
- [ ] Number format (more of a bug than a feature)
- [ ] Colorized bar and buttons
- [ ] Percentage display without clicking or hovering

I don't know how to do this in userscript -- perhaps store options as a constant?

cyrildtm:
I noticed the addition of local/synced storage and advanced settings page. Ok I will try and spend some spare time to implement at least some of those I have mentioned.

ErykDarnowski:
Maybe a dict with would be a nicer option to store that data compared to constants. 🤔

sy-b:
👍

cyrildtm:
Maybe, but depends. Since the function returning color strings is called once for every video, and once every time you select a new scheme in the popup, and since we don't have a million colors to store, there is no real performance gain. Readability does not get improved either, since you will need to keep a separate paragraph to actually write down the colors. There is little to no storage space save; noting that popup and content script must absolutely not share anything per security requirements. 

One possible way to reduce eye strain: 
https://github.com/cyrildtm/return-youtube-dislike/blob/f218443b4c8799491a6662e5ee2be0057c56d40b/Extensions/combined/src/utils.js#L136

```
    case 'classic':
    default:
       colorString = voteIsLike ? 'lime' : 'red';
```

But we are considering light/dark theme adaptive colors at this moment, so this may not help at all.

PR #435, opened 2022-01-11
Add custom number formats

### LINKED ISSUE #201: Format dislikes number the same way likes are formatted

### LINKED ISSUE #364: (Feature Request): Only display an approximate value of dislikes if YouTube also returns likes as an approximate value.
### Extension or Userscript?

Extension

I don't know if I'm explaining this right, but here's the thing: for me, dislikes are displayed as "1,7 MIL" (as an example), while likes are displayed as an "exact value" (150.749).

Here's an example:
image

I can even see the exact amount of dislikes that the video has, so it shouldn't be that hard to implement it:
image

In case this couldn't be done; the extension could also modify the like count to an approximate value (just like the dislike count) if the user wanted to, but I don't see why wouldn't it be possible.

### Ways to implement this!

A way to implement this is checking if the like count is an exact value or an approximate one. Then, change the dislike button based to a different value using the like button.

### Can you work on this?

- [ ] Yes
- [X] No

### LINKED ISSUE #407: (Bug): Localized large dislike numbers are incorrect for some languages
### Extension or Userscript?

Both

For all dislikes greater than 999 and certain non-English languages like Thai, RYD only formats them in K, M and B.

image

They look apart. Note that "5.8 หมื่น" (Thai) translates to "5.8 ten-thousands" which is 58000. As for dislikes, it is "1.3 พัน" in Thai (1.3 thousands).

### Ways to implement this!

You will write it like this for Thai locale:

```js
new Intl.NumberFormat('th-TH', {
  notation: "compact",
  compactDisplay: "long"
}).format(58000); // return 5.8 หมื่น
```

We may need a mapping between locale and `compactDisplay` value?

Or, do this?

```js
const formattedEn = new Intl.NumberFormat('en-US', { notation: "compact" }).format(dislikes);
let result = new Intl.NumberFormat(targetLocale, { notation: "compact" }).format(dislikes);
if (formattedEn.localeCompare(formattedLocalized) != 0) {
    result = new Intl.NumberFormat(targetLocale, { notation: "compact", compactDisplay: "long" }).format(dislikes);
}
```

### Can you work on this?

- [X] Yes (but it may be somewhat messy)
- [ ] No

cyrildtm:
Resolve #201, resolve #364, resolve #407
Using plugin options (#387)


cyrildtm:
Certain locales may display like number without rounding down. See: issue #364
```
Intl.NumberFormat('de', {notation: 'compact', compactDisplay: 'short'}).format(123456)
'123.456'
```

Extensions/combined/popup.js
```diff
@@ -94,13 +135,61 @@ function storageChangeHandler(changes, area) {
...
+}
+
+function handleNumberDisplayFormatChangeEvent(value) {
+  config.numberDisplayFormat = value;
+  document
+    .getElementById("number_format")
+    .querySelector('option[value="' + value + '"]').selected = true;
+}
+
+function getNumberFormatter(optionSelect) {
```
Anarios, line 165: same function down below?

Anarios (changes requested):
Some code is copy-pasted, is it possbile to get rid of duplicates?

Extensions/combined/popup.js
cyrildtm: It provides the same function but at a different place. The main feature is in util.js that works in the user's window (youtube.com) / content script; while this one in popup.js works in the extension's popup interface. 
I added this function (and others alike) to show an example of what the user really sees using their locale settings. 

I tried to import these functions from util.js or other files into popup.js and vice versa, but got errors.

cyrildtm: Otherwise there's a code-sharing scheme [like this one](https://stackoverflow.com/a/24227952/3839249) but that looks as messy as this one. Let me know if there's any other way of implementing one function in both places at once, and I can make the change.

Anarios:
Thanks a lot

PR #460, opened 2022-01-23
bring OpenGraph text in line with current project status

SushiInYourFace:


SushiInYourFace:
Nothing major, just noticed the opengraph text was outdated

Anarios (approved):

PR #475, opened 2022-02-10
Added Kiwi Browser compatibility info in installing guide

sy-b:
This was added approximately a month ago in the wiki.


Anarios:
Thanks!

PR #493, opened 2022-02-25
Adding Dislike Button to Comments + Adding Percentage of Likes/Dislikes

qwewas123:


sy-b (commented):
Tl;dr Wouldn't work

---

Reasons
- It's simply a file lying around in the main folder. It wouldn't be included in the build process.
- It's mostly copy-pasted and adapted to include comment dislikes **without changing sendVote()**
- sendVote() doesn't work that way. Simply doing sendVote(1) even on comments would be interpreted as a vote to the video. The backend wouldn't accept multiple likes/dislikes from the same userID.
-  To submit comment ratings, either sendVote() needs to be modified or another function must be made.
- The API can only accept like/dislike for video & not for comments.

sy-b:
It's infeasible 

- Each comment would need its own entry into the database.
- Even if only the top comments are stored, the database would still be enormous.

Issue #472, opened 2022-02-03
(Bug): "DISLIKE DISABLED" on new Videos


agent-LuluDodo:
### Browser

Firefox

### Browser Version

Firefox 97

### Extension or Userscript?

Extension

### Extension/Userscript Version

v2.1.0.3

### Video link where you see the problem

The Video needs to be under 5 min old

### What happened?

On Videos which are new, the Dislike count shows as "Dislikes Disabled".
Shouldn't that be more like "Dislikes calculating", "Not enough info" or "Dislike Unknown" _(with a tooltip aside explaining possible reasons)[credits for this one: @cyrildtm]_?

![Dislike Disabled](https://user-images.githubusercontent.com/98981925/152416856-14901f9d-a083-4158-b0ef-8704a82b9170.png)

cyrildtm:
possible duplicate of #461
Please see if your problem resolves by itself in a while

cyrildtm:
possible duplicate of #444 
Please see if your problem resolves by itself if you like the video once (and refresh) or wait for a while

andry81:
I know the video from 2022.01.31 and it still has it disabled. I think it is not about when it has published.

sy-b:
"Dislike Disabled" means that the creator has disabled both likes & dislikes.

The extension uses the number of likes to extrapolate the dislikes count.

Without like count it can't extrapolate

agent-LuluDodo:
> "Dislike Disabled" means that the creator has disabled both likes & dislikes.
I'm the creator of this Video and I have likes and dislike enabled.

I understand that there is not enough info to display the dislike count, but I think the message "Dislike Disabled" is misleading. It should be more like "Dislikes calculating" or "Not enough info"

agent-LuluDodo:
> I know the video from 2022.01.31 and it still has it disabled.

Send the Link!

agent-LuluDodo:
> possible duplicate of #444 Please see if your problem resolves by itself if you like the video once (and refresh) or wait for a while

Yes, it does show up after a while. But the Bug report is to show that this is misleading. It should be something like "Dislikes calculating" or "Not enough info".

cyrildtm:
What about saying "Dislike Unknown" and put a tooltip aside explaining possible reasons

agent-LuluDodo:
> What about saying "Dislike Unknown" and put a tooltip aside explaining possible reasons

That sounds better than mine. Adding this as a possibility in the Description.

[...]

PR #527, opened 2022-04-06
Display 'temporarily unavailable' when the API can't be reached

CreaZyp154:
Display 'temporarily unavailable' instead of the dislike count when the API can't be reached.

I also made a new function in `utils.js` for localizing strings from the files in `_locales` and already translated it in french because it is my native language.

### DIFF
--- a/Extensions/combined/src/state.js
+++ b/Extensions/combined/src/state.js
@@ -8,6 +8,7 @@ import {
   getColorFromTheme,
 } from "./utils";
 import { sendVideoIds } from "./events";
+import { localize } from "./utils";
 
 //TODO: Do not duplicate here and in ryd.background.js
 const apiUrl = "https://returnyoutubedislikeapi.com";
@@ -124,6 +125,13 @@ function processResponse(response, storedData) {
   }
 }
 
+// Tells the user if the API is down
+function displayError(error) {
+  getButtons().children[1].querySelector("#text").innerText = localize(
+    "textTempUnavailable"
+  );
+}
+
 async function setState(storedData) {
   storedData.previousState = isVideoDisliked()
     ? DISLIKED_STATE
@@ -144,8 +152,12 @@ async function setState(storedData) {
       },
     }
   )
+    .then((response) => {
+      if (!response.ok) displayError(response.error);
+      return response;
+    })
     .then((response) => response.json())
-    .catch();
+    .catch(displayError);
   cLog("response from api:");
   cLog(JSON.stringify(response));
   likesDisabledState =
--- a/Extensions/combined/_locales/en/messages.json
+++ b/Extensions/combined/_locales/en/messages.json
@@ -29,6 +29,9 @@
     "textSettingsHover": {
         "message": "Stops counting your likes and dislikes."
     },
+    "textTempUnavailable": {
+        "message": "temporarily unavailable"
+    },
     "textUpdate": {
         "message": "update to"
     },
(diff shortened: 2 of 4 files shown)


cyrildtm:
Hi @CreaZyp154 Could you take a look at #472 ?

CreaZyp154:
@cyrildtm I have now fixed the issue

Extensions/combined/src/bar.js
```diff
@@ -61,7 +62,9 @@ function createRateBar(likes, dislikes) {
   } else {
     cLog("removing bar");
     let ratebar = document.getElementById("ryd-bar-container");
-    ratebar.parentNode.removeChild(ratebar);
+    if(ratebar) {
```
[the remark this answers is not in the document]
CreaZyp154: I got errors when the video's likes count was disabled it sayed cannot read properties of undefined and pointed to ratebar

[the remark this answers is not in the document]
CreaZyp154: so I'm checking if it exists first

Extensions/combined/src/state.js
```diff
@@ -42,6 +39,16 @@ function isShorts() {
...
 }
 
+function isLikesDisabled() {
+  // return true if the like button's text doesn't contain any number
+  if (isMobile()) {
+    return /^\D*$/.test(
+      getButtons().children[0].querySelector(".button-renderer-text").innerText
+    );
+  }
+  return /^\D*$/.test(getButtons().children[0].querySelector("#text").innerText);
```
cyrildtm, line 49: Is it ever possible that, does such language ever exist in the world that, a shortened number starts with a non-Arabic number digit? Just like we put Euro or Dollar signs in the front.

CreaZyp154: Wow, I never thought of that, I think they're other numbering systems (not 100% sure) so I would have to update that regex

Extensions/combined/src/bar.js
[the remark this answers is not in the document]
CreaZyp154: Oh ok yeah I mean null is falsey so it works. idk if it's good practice but it's pretty clear imo

Extensions/combined/src/state.js
CreaZyp154: Well that would be a [big regex](https://tc39.es/ecma402/#table-numbering-system-digits) and ik that while there's a lot of numeral systems, most languages use the western arabic one (0-9)

cyrildtm: No sorry I didn't mean a crazy cover-all utility book. I think as long as there is a count, there will be number digits, and otherwise none for the opposite case. `\d+` should suffice

P.S. Arabic is not western :P lol surprise

CreaZyp154: Ok then should i close that ?

cyrildtm: Maybe with `\d+`? I would start with the least strict pattern first anyway. Hope Youtube won't come up with some weird formats down the road.

CreaZyp154: why inverting and why + ?

cyrildtm: Sorry I was a bit confused. Any non-digit from beginning to end and at least one digit anywhere should produce the same result, unless the field is empty.

cyrildtm:
Hey @CreaZyp154  sorry to bother you again, but would you please take a look at #538 ? Seems to be another exceptional case. In the referred video, Youtube disabled the like counter after the video is posted.

PR #535, opened 2022-04-16
Removed popup on update/install

### LINKED ISSUE #534: (Feature Request): Don't pop up changelog on update
### Extension or Userscript?

Extension

I find it unacceptable when an extension pops open a tab without prompting (to the point where, in the past, I filed a request with addons.mozilla.org that their policy should be amended to make it a violation of their terms unless it's an opt-in (i.e. off-by-default) option) and don't want to have to put up with it.

If you are unwilling to at least make a highly visible option to disable it, please let me know promptly so I can remove Return YouTube Dislike from my browser... and probably my mother's too.

When HTTPS Everywhere was doing this sort of thing, it would prompt her to call me, fearful that she'd been hacked.

### Ways to implement this!

Either remove the on-update changelog displaying entirely or add a highly visible toggle in said popup to allow people to enable or disable it.

### Can you work on this?

- [ ] Yes
- [X] No

sy-b:
Popup on update/install isn't really necessary unless there's a major & important announcement to make. We can add it later, when it is actually needed.

resolves #534

### DIFF
--- a/Extensions/combined/manifest-chrome.json
+++ b/Extensions/combined/manifest-chrome.json
@@ -15,9 +15,6 @@
   "permissions": [
     "storage"
   ],
-  "action": {
-    "default_popup": "popup.html"
-  },
   "content_scripts": [
     {
       "matches": [


sy-b:
resolves #79

sy-b:
oops!

fixing it

Anarios:
Do not remove it completely, there will be an opt-out option

sy-b:
> there will be an opt-out option

This is a one time pop up & we aren't pushing updates frequently. Right?

Anarios:
The thing that you removed now is the settings window that appears when you click on extension icon

sy-b:
> The thing that you removed now is the settings window that appears when you click on extension icon

yeah, I had to remove this 😅
```JS
api.runtime.onInstalled.addListener(() => {
  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
})
```

sy-b:
I was just waiting for you response on that opt-out option thing. 🙂

Anarios:
> This is a one time pop up & we aren't pushing updates frequently. Right?

We still want to show it to new users, so I still wouldn't want to just remove it, an opt-out option is better.

sy-b:
Ok, got it 👍

PR #537, opened 2022-04-16
Show changelog only on installation

### LINKED ISSUE #534: (Feature Request): Don't pop up changelog on update

sy-b:
fixes #534

### DIFF
--- a/Extensions/combined/ryd.background.js
+++ b/Extensions/combined/ryd.background.js
@@ -74,9 +74,12 @@ api.runtime.onMessage.addListener((request, sender, sendResponse) => {
   }
 });
 
-api.runtime.onInstalled.addListener(() => {
-  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
-})
+api.storage.sync.get(['newInstallation'], (result) => {
+  if (result.newInstallation !== false) {
+    api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
+  }
+});
+api.storage.sync.set({'newInstallation': false});
 
 async function sendVote(videoId, vote) {
   api.storage.sync.get(null, async (storageResult) => {


cyrildtm:
how to prevent popup forcefully showing up for the second time?

sy-b:
<div type='discussions-op-text'>

<sup>Originally posted by in #556, April 13, 2022</sup>
Another problem -

In future we might have the need to force the popup regardless of user's choice.

For e.g. when
1. signing up is made compulsory for vote submission to prevent API abuse (I hope we never have to do this)
2. we add features like multi device support (to prevent multiple dislike submissions)
3. creator dislike submission
4. Name change (unlikely) (refer: #509)

Here, we'll (in #556) have to comment out the code temporarily. My idea (in #537) behind using stored value was that we can override it programmatically from somewhere. I still don't have any plans on it's implementation to do this but it just seemed better to me than asking the browser & then adding conditions on it. Of course, if there's no good reason to use a stored value then this certainly a good way. 🙂

</div>

sy-b:
> how to prevent popup forcefully showing up for the second time?

I didn't get the question. Could you elaborate?

sy-b:
You mean that you want to prevent it from showing in next update?

sy-b:
In #541, you mentioned
> Default is disabled, which means you won't see updates.

Shouldn't this solve it?

sy-b:
> I still don't have any plans on it's implementation to do this but it just seemed better to me than asking the browser & then adding conditions on it.

I think we should include `overrideDisabledPopup` variable somewhere.

Extensions/combined/ryd.background.js
```diff
@@ -74,9 +74,12 @@ api.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
 
-api.runtime.onInstalled.addListener(() => {
-  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
-})
+api.storage.sync.get(['newInstallation'], (result) => {
+  if (result.newInstallation !== false) {
+    api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
+  }
+});
+api.storage.sync.set({'newInstallation': false});
```
bershanskiy, line 82: I'm sorry to mention this, but there is a race condition between two async calls. Could you please move `api.storage.sync.set` call into callback of `api.storage.sync.get`?

```
api.storage.sync.get(['newInstallation'], (result) => {
  if (result.newInstallation !== false) {
    api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
    api.storage.sync.set({'newInstallation': false});
  }
});
```

cyrildtm:
> In #541, you mentioned
> 
> > Default is disabled, which means you won't see updates.
> 
> Shouldn't #542 resolve that problem

The opt-in logic is not "yes means yes". It is OR-ed with new install event. 

https://github.com/Anarios/return-youtube-dislike/blob/b49bca3f24f0b623b085f64926196843513112b7/Extensions/combined/ryd.background.js#L78

Extensions/combined/ryd.background.js
cyrildtm: This is better. Another reason is we don't need to rewrite it every time a page is open.

cyrildtm:
> > I still don't have any plans on it's implementation to do this but it just seemed better to me than asking the browser & then adding conditions on it.
> 
> I think we should include `overrideDisabledPopup` variable somewhere.

That sounds like a big brother move. No, please.

sy-b:
> That sounds like a big brother move
😂

> No, please.
👍

If we are not going to override user's choice, then #555 is better.

Extensions/combined/ryd.background.js
sy-b: For those who are going to receive updates, `newInstallation` will never be set to `false`.

Now this is getting problematic.

If we are not going to override user's choice, then #555 is better.

cyrildtm:
> If we are not going to override user's choice, then #556 is better.

But that one shows a new tab for every extension update, and it is not compatible with opt-in. 

Can we open up a Issue to collaborate? There are too many conditions to check and we can do it better. 
Keep the current PR open, so we can cherry pick, as soon as we get a consensus and open up a new PR.

Extensions/combined/ryd.background.js
cyrildtm: @sy-b again, #556 is not compatible with opt-in. Also, for those who want to see changelog, `newInstallation` has no effect, and we can treat it as a don't-care.

sy-b:
> There are too many conditions to check

```
show changelog
  if (extension is updated && user has opted in)
```

Am I missing something?

cyrildtm:
#556   filter out non-extension update events
```
api.runtime.onInstalled.addListener((details) => {
  if (
    // No need to show changelog if its was a browser update (and not extension update)
    details.reason === "browser_update" ||
    // No need to show changelog if developer just reloaded the extension
    (details.reason === "update" &&
      details.previousVersion === chrome.runtime.getManifest().version)
  )
    return;
  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
})
```

#541 (based on old #537)  user opt-in option
```
api.storage.sync.get(['newInstallation'], (result) => {
  if (result.newInstallation !== false 
    || extConfig.showUpdatePopup === true
  ) {
    api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
  }
});
api.storage.sync.set({'newInstallation': false});
```

#537 (new) Show for a second time and no more
```
api.storage.sync.get(['showChangelog'], (details) => {
  if (details.showChangelog !== false) {
    api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
    api.storage.sync.set({'showChangelog': false});
  }
});
```




> show changelog
>   if (extension is updated && user has opted in)


```
api.runtime.onInstalled.addListener((details) => {
  if (
    // No need to show changelog if its was a browser update (and not extension update)
    details.reason === "browser_update" ||
    // No need to show changelog if developer just reloaded the extension
    (details.reason === "update" &&
      details.previousVersion === chrome.runtime.getManifest().version)
  ) {
    return;
  }
  api.storage.sync.get(['showChangelog'], (result) => {
    if (result.showChangelog !== false 
      || extConfig.showUpdatePopup === true) {
      api.storage.sync.set({'showChangelog': false}); // keep it inside get to avoid race condition
      api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")}); // wait until async get runs
    }
  });
})
```

sy-b:
Is `details.reason === "browser_update"` necessary?
After a browser update, either the extension version will remain the same or be changed. This is handled by the next conditions. Do we still need it?

Is this fine?

```JS
api.runtime.onInstalled.addListener((details) => {
  if (
    details.reason === "update" &&
    details.previousVersion === chrome.runtime.getManifest().version
    ) {
    return;
  }
  api.storage.sync.get(['showChangelog'], (details) => {
    if (details.showChangelog !== false ||
      extConfig.showUpdatePopup === true) {
      api.storage.sync.set({'showChangelog': false});
      api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html"), active: false});
    }
  });
});
```

cyrildtm:
No. 

https://developer.chrome.com/docs/extensions/reference/runtime/#event-onInstalled

> Indicates the previous version of the extension, which has just been updated. This is present only if 'reason' is 'update'.

> 
> The reason that this event is being dispatched.
> 
> TYPE
> 
> "install"
> 
> ,
> "update"
> 
> ,
> "chrome_update"
> 
> , or
> "shared_module_update"


At least we need to separate browser update events. 


Besides, does `chrome.runtime` exist in Firefox?

cyrildtm:
even better: 

```
  api.storage.sync.get(['lastShowChangelogVersion'], (details) => {
    if (extConfig.showUpdatePopup === true &&
      details.lastShowChangelogVersion !== chrome.runtime.getManifest().version
      ) {
      // keep it inside get to avoid race condition
      api.storage.sync.set({'lastShowChangelogVersion ': chrome.runtime.getManifest().version});
      // wait until async get runs & don't steal tab focus
      api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html"), active: false});
    }
  });
```

- Saves me from having my focus stolen seven times from my seven computers
- Prevents the second time popup before opt-in

sy-b:
> Besides, does `chrome.runtime` exist in Firefox?

Yes

sy-b:
Should I close this PR?

cyrildtm:
Maybe, if you want to open a new one. But maybe we can just keep using this, and keep others until this one gets merged.

cyrildtm:
I'm really sorry for being lazy. I didn't mean to remove the outer event listener part, since we still filter out non-extenstion updates

sy-b:
>  since we still need to filter out non-extension updates

`onInstalled` will be triggered after a -
- Browser Update
- Extension Update
- Extension Reload
- Extension Install

Changelog must be shown only when the extension has actually updates.

It can be done by
1. filtering out others
**or**
2. by checking for version number change.

If we are just checking for version number change, then why is there a need to check `details.reason === "browser_update"` ?

cyrildtm:
Rare case. Will elaborate later.

cyrildtm:
The numbers mean sequentially happened events

Rare case #  1: without event listener or install reason check
1. Extension updates. 
2. User notices opt-in option and switches on. Then closes browser. 
3. User opens youtube or whatever tab that has extension background script called. 
4. User receives an outdated update popup. 

Rare case #  2: with event listener but without install reason check
1. Extension updates. 
2. User notices opt-in option and switches on. Then closes browser. 
3. Browser updates.
4. User opens youtube or whatever tab that has extension background script called. 
5. User receives an outdated update popup. 

Previously I was also concerned about `details.previousVersion` being undefined for other reasons than extension update, so that in your old comment `if (details.previousVersion === chrome.runtime.getManifest().version)` does not proceed as true. But we are no longer using this method, and apparently `details` variable name got re-used.

sy-b:
\:facepalm  I don't know why I mindlessly assumed `details.lastShowChangelogVersion` = `onInstalled` & also as a stored variable.

In case 2, are you worried about `details.lastShowChangelogVersion` being `undefined`/undeclared?
That can only happen if browser update clears all the data and eventually `'lastShowChangelogVersion'` & `extConfig.showUpdatePopup` will also be `undefined`/undeclared. So nothing should happen. Also, there shouldn't be a need for adding an event listener.

Going by this logic, I was not able to understand the reason for case 1.
Case 1 can happen if ryd.background.js doesn't immediately run after any install. But it does.

cyrildtm:
without the event listener, the inner logic runs every time the background script is loaded; a browser restart ensures it reloads. 

`extConfig` gets initialized as soon as background script is loaded. However `lastShowChangelogVersion` does not get its value until the first time a changelog is shown in a popup. There is a time gap between when these two things happen. 

I wonder if there is a way to force trigger a browser update event. When I get some time I can verify if I can manually get into these rare cases. 

Based on this extension's Chrome store one-star reviews (many and recent), getting "donation" popup multiple times is not unheard of. Unless they are bots, of course.

sy-b:
> `lastShowChangelogVersion` does not get its value until the first time a changelog is shown in a popup.

I feel that this is ok. Just add a huge slider button asking "Inform about future updates?" in changelog.html. (Remove this later)
- This might generate some backlash but not much a it as won't steal focus & will be asking users if they want see this in future.
- Good for those who want to opt-in.
But its still fine if it doesn't. My point is that not all users are annoyed. Discord was _mostly_ flooded with people asking - how to get to the settings menu.

<br>

> Based on this extension's Chrome store one-star reviews (many and recent), getting "donation" popup multiple times is not unheard of. Unless they are bots, of course.

Hmmm...  🆗 👍

cyrildtm:
Yeah it's all about how much you care about that one percent of people. We are the one percent who chose to use the extension, don't forget.

cyrildtm:
If update popup opt-in is used, then this PR should cherry pick #541, which cherry picked #537 .... which should be fine

Found a potential problem: 
```
  api.storage.sync.get(['lastShowChangelogVersion'], (details) => {
    if (extConfig.showUpdatePopup === true &&
```
In background script, `extConfig` gets updated with stored values from `initExtConfig();` However, these are async get methods. If we use another async method right after this, there may be a race condition where update opt-in value is not updated to reflect user's choice. The default value `false` may be used instead, and popup may never show up. 

In fact, `extConfig.showUpdatePopup` would be the first ever use of user options in background script, which might be dangerous. The service worker is supposed to initialize or load user options instead of using them. 

One workaround is to embed the popup logic inside `initializeShowUpdatePopup()`

cyrildtm:
> embed the popup logic inside `initializeShowUpdatePopup()`

But in this way I can still see the old changelog ....

Clarification: Suppose v3.1 introduces opt-in. A user gets extension update with Chrome 100, finds the opt-in switch immediately, and turns it on. The next day, Chrome gets updated to 100.1, and a popup shows v3.1 changelog, which is "old" by then.

Issue #539, opened 2022-04-16
(Feature Request): New user options


cyrildtm:
### Extension or Userscript?

Both

### Request or suggest a new feature!

This is a bookkeeping thread to track my plan in the near future. I can take care of these issues: 

#520 userscript locale number format
- [ ] add userscript user option
- [ ] add number format tuner
- [ ] add optin for vote submission
- [ ] add colorized thumbs

#535 no update popup by default
- [ ] add extension user option
- [ ] add userscript option

It's mainly porting implemented functionalities over, but also a bit new stuff

### Ways to implement this!

Introduce user options to userscript. Add option to opt in update popup. 

### Can you work on this?

- [X] Yes
- [ ] No

cyrildtm:
Vote submission and update popup are not implemented in Userscript at this moment. Will be able to continue if they are to be added in the future.

sy-b:
For update popup -

To update, users are manually going to update it, so, update page can be shown on installation.

cyrildtm:
@sy-b  you mean for the userscript?

PR #540, opened 2022-04-18
User script user options

### LINKED ISSUE #520: (Bug): Dislikes count on the userscript are not correct on some locales
### Browser

Firefox, Vivaldi

### Browser Version

Firefox Stable 98.0.2, Vivaldi 5.1.2567.66 (Stable channel)

### Extension or Userscript?

Userscript

0.9.0

https://www.youtube.com/watch?v=VgfIxCsg8xY

### What happened?

The dislike returned by the api was 1412, but the dislikes count showed 1,4M.

The issue seems to be on the numberFormat function of the userscript, most specifically the locale part of it, since just replacing all the contents of the function with this makes it work as it should:

```
function numberFormat(numberState) {
  return numberState.toLocaleString();
}
```

That is without using the roundDown function, but using it works all the same, I just prefered to see the actual number instead of the rounding.

Install userscript on violentmonkey, use spanish locale, and go to the video link of the problem.
Happens on both Firefox and Vivaldi (Vivaldi uses the rendering engine of chrome 98 at the point of testing)

cyrildtm:
introduce user options to userscript. Using existing options available in extension for now, plus showing update popup.
Resolves #520 
Not implemented features but with options: disable vote submission, show update popup. 

Part of #539

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -25,8 +25,8 @@
 
 const extConfig = {
 // BEGIN USER OPTIONS
-// You may change the following variables to allowed values listed in the bracket (* means default). Keep the style and keywords intact. 
-  showUpdatePopup: false, // [true, false*] Show popup tab after extension update (See what's new)
+// You may change the following variables to allowed values listed in the corresponding brackets (* means default). Keep the style and keywords intact. 
+  showUpdatePopup: false, // [true, false*] Show a popup tab after extension update (See what's new)
   disableVoteSubmission: false, // [true, false*] Disable like/dislike submission (Stops counting your likes and dislikes)
   coloredThumbs: false, // [true, false*] Colorize thumbs (Use custom colors for thumb icons)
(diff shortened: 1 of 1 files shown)


ErykDarnowski (approved):
Nicely done! Besides the first 2 values in extConfig (showUpdatePopup and disableVoteSubmission) that aren't implemented yet, everything looks great ;)

Extensions/UserScript/Return Youtube Dislike.user.js
```diff
@@ -347,14 +379,73 @@ function numberFormat(numberState) {
...
+      }
+      break;
+    case "neon":
+      if (voteIsLike === true) {
+        colorString = "aqua";
+      } else {
+        colorString = "magenta";
+      }
+      break;
+    case "classic":
```
ErykDarnowski, line 440: Hey I'm wondering why did you use `case "classic":` here when there's already `default:` is it for readability?

cyrildtm: clarity.

PR #541, opened 2022-04-18
Add extension opt-in option to show update popup

cyrildtm:
Add an opt-in option to display a new tab for changelog after each update. 
Default is disabled, which means you won't see updates. 
This comes after PR #537 with a temporary cherry pick. Tell me to rebase once that one is merged. 
Improves #534 
Part of #539


cyrildtm:
Basically....

Capture

Anarios:
3.0.1 is submited to the store for review.

Anarios:
@cyrildtm, do you use chrome canary?

cyrildtm:
> @cyrildtm, do you use chrome canary?

Yes I'm on 103.0.5027.0
Now I'm making one more fix. Will explain in a second.

cyrildtm:
An event listener for onInstalled is still a better solution. Here's an simple experiment: 

in background js, add `api.tabs.create({url: "about:blank"});` at the top level. 

Wait until the service worker goes "inactive"; This can be easily observed when you have chrome://extensions/ open, while reading through [this page](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/) explaining why service workers get terminated frequently. 

Then, open extension settings and flip a switch. You will see a new tab. 

Conclusion, background js (or service worker, whatever you call it) gets re-run frequently. It's quite a waste to read options. And why.

cyrildtm:
And .... 
https://github.com/Anarios/return-youtube-dislike/blob/2c39c096be468954ba30d9c4fb7bc55f789e020d/Extensions/combined/ryd.background.js#L81

This version will never get a popup 😃

Anarios:
> And ....
> 
> https://github.com/Anarios/return-youtube-dislike/blob/2c39c096be468954ba30d9c4fb7bc55f789e020d/Extensions/combined/ryd.background.js#L81
> 
> This version will never get a popup 😃

It gets a popup only on install, just as intended for now.

Anarios:
` It's quite a waste to read options.`
Is it an expensive operation, though? I thought that it costs almost nothing.

cyrildtm:
> It gets a popup only on install, just as intended for now.

I just noticed your words "for now". Okay, but we will need it to actually work in the future right? 


> it costs almost nothing.

A non-zero cost action is not nothing. 
Currently we have seven options and therefore seven async gets to read the options, separately. One can justify "separately" with text clarity and async gets bunched anyway. This little feature causes another two async reads, which is +28%. Do it two million times. Not because we sacrifice efficiency for tidiness, but simply because we can. So yes, it is a waste. 

Service worker gets killed and reinstated quite often, it's not like once every browser update. That's my understanding.

Anarios:
> Okay, but we will need it to actually work in the future right?

If there are ever changes big enough that everyone needs to be notified - then yeah. Right now I thought that showing it only on install is sufficient. I wanted to release the fix ASAP to not annoy people with Canary, so I went with the simplest solution of just disabling it.

> This little feature causes another two async reads

Can be refactored into a single init function that reads all settings in one go.

> Service worker gets killed and reinstated quite often

Every few minutes, if I'm not wrong.

cyrildtm:
Refactoring can solve everything together.

Anarios:
I believe there were no more complaints on popups, I will be closing this for now.

PR #556, opened 2022-04-20
Fix: do not open changelog if not needed

### LINKED ISSUE #555: (Bug): Extension opens changelog URL after browser update
### Browser

Google Chrome

### Browser Version

100

### Extension or Userscript?

Extension

3.0.0.0

Any

### What happened?

Browser updated and extension opened v3 Changelog. This occurs because `onInstalled` handler does not check [install reason](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/OnInstalledReason):

https://github.com/Anarios/return-youtube-dislike/blob/2e6ed5298b7851b7ea84dcfa298167689f516496/Extensions/combined/ryd.background.js#L77-L79

Wait for browser update and start browser, see that extension opened a new tab. If you use Canary, this update will happen every day.

bershanskiy:
Fixes #555 by checking extension installation details.

### DIFF
--- a/Extensions/combined/ryd.background.js
+++ b/Extensions/combined/ryd.background.js
@@ -74,9 +74,19 @@ api.runtime.onMessage.addListener((request, sender, sendResponse) => {
   }
 });
 
-api.runtime.onInstalled.addListener(() => {
-  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
-})
+api.runtime.onInstalled.addListener((details) => {
+  if (
+    // No need to show changelog if its was a browser update (and not extension update)
+    details.reason === "browser_update" ||
+    // No need to show changelog if developer just reloaded the extension
+    (details.reason === "update" &&
+      details.previousVersion === chrome.runtime.getManifest().version)
+  )
+    return;
+  api.tabs.create({
+    url: api.runtime.getURL("/changelog/3/changelog_3.0.html"),
+  });
+});
 
 async function sendVote(videoId, vote) {
   api.storage.sync.get(null, async (storageResult) => {


Extensions/combined/ryd.background.js
```diff
@@ -74,9 +74,19 @@ api.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
 
-api.runtime.onInstalled.addListener(() => {
-  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
-})
+api.runtime.onInstalled.addListener((details) => {
+  if (
+    // No need to show changelog if its was a browser update (and not extension update)
+    details.reason === "browser_update" ||
+    // No need to show changelog if developer just reloaded the extension
+    (details.reason === "update" &&
```
bershanskiy, line 82: This is for when developer clicks "reload" button on `chrome://extensions/`.

Extensions/combined/ryd.background.js
```diff
@@ -74,9 +74,19 @@ api.runtime.onMessage.addListener((request, sender, sendResponse) => {
   }
 });
 
-api.runtime.onInstalled.addListener(() => {
-  api.tabs.create({url: api.runtime.getURL("/changelog/3/changelog_3.0.html")});
-})
+api.runtime.onInstalled.addListener((details) => {
+  if (
+    // No need to show changelog if its was a browser update (and not extension update)
+    details.reason === "browser_update" ||
```
bershanskiy, line 80: This is the actual fix for #555.

cyrildtm:
how is this different than #537 and #541?

sy-b:
Another problem -

In future we might have the need to force the popup regardless of user's choice.

For e.g. when
1. signing up is made compulsory for vote submission to prevent API abuse (I hope we never have to do this)
2. we add features like multi device support (to prevent multiple dislike submissions)
3. creator dislike submission
4. Name change (unlikely) (refer: #509)

Here, we'll have to comment out the code temporarily. My idea (in #537) behind using stored value was that we can override it programmatically from somewhere. I still don't have any plans on it's implementation we do this but it just seemed better to me than asking the browser & then adding conditions to it. Of course, if there's no good reason to use a stored value then this certainly a good way. 🙂

bershanskiy:
If #537 or #541 get rid of `api.runtime.onInstalled`, then this PR is redundant.

> In future we might have the need to force the popup regardless of user's choice.

You mean force opening changelog in a new tab (not popup)? Also, (in my personal opinion) I oppose anything that goes against explicit user choice.

sy-b:
> anything that goes against explicit user choice

I understand. But just sometimes in some critical cases, we might need to.
For example - 
You can disable all promotional emails from GitHub but not the critical ones like account recovery, privacy policy updates & other legal or administrative emails.

I view this (RYD's) option just like that. Yes, the user can disable almost all the notifications/popups (i.e. changelogs in the new tab) from RYD, but the critical ones. I don't know when we'll need this, and we can work on overriding when it's actually needed.
I just brought up the point now to ease the future work.

cyrildtm:
> critical ones

To be honest, being able to paint the thumbs with colors is nice and great, but if I don't know there's this feature, I'm not disappointed. I would like the surprise of discovering new features **by myself** every now and then. 

But I don't want a new tab open every time there's an incremental update. A typo gets fixed, a link is updated, there's a new line between this sentence and the next. I don't give a damn, and I don't need you blasting this bubblegum to my face. 

For this fun project, I think the only critical message is our final goodbye, either when youtube decides to return dislikes itself, or we get a C&D. In either case, just use the simplest way and open a tab. Uninstall. Done. Move on with life.

sy-b:
👍

Issue #211, opened 2021-12-05
Option to show like/dislike percentage instead of full like/dislike count


Nightcaat:
![image](https://user-images.githubusercontent.com/26909213/144755140-44cc257b-4071-47a0-915a-fae307cbe23a.png)

SirLich:
If... this does make it into a feature (I'm actually somewhat in favor), I would rather see it represented as a single number (like percentage), not two. 

Additionally... isn't that what the 'bar' is for?

magnus-ISU:
That's what the bar is for, but its nice to have the actual numerical proportion rather than a vague line.

Agree a single number would be nicer.

ErykDarnowski:
I think it would be nice to give users a couple options of displaying this info:
- 190 / 10 - 95%
- 95% / 5%
- 95%

ErykDarnowski:
Got a dirty implementaion of this ^ done already, I'll work on this more tomorrow.

Nightcaat:
Good stuff :]

I got a notification here and cloned your fork but I couldn’t get it to work on either browser. It looks like you re-forked the repository so I assume you’re still working on it?

ErykDarnowski:
It's actually fully implemented already! After I fix this last bug I'm pushing it to my fork ;)

ErykDarnowski:
It's done :)

Nightcaat:
Oops, forgot to close my own issue. Fixed in #560

Issue #221, opened 2021-12-06
(Feature Request): Show accuracy in some way


KeinDahLahzey:
### Extension or Userscript?

Both

### Request or suggest a new feature!

I suggested this in the discord but felt that it might be good to add here so its not lost/fully explained in one go. 

Anyhow a system of some kind to display how accurate the dislike count is once the API closes off. 

### Ways to implement this!

There are a few ways to implement on the front end but the way i immediately thought of was a colored tier system.

Tier 1: Creator has shared dislike count.
Tier 2: Dislikes were archived and there is extension user data to adequately update.
Tier 3: Dislikes were not archived but enough extension users to sample data for accurate dislike count.
Tier 4 : Not enough extension users to sample reliable data from.

To display this the like number would be changed to a different color to indicate the reliability ranging from green at tier 1 to red at tier 4. 

Another idea would be to have 'badges' beside the dislikes to display this, and hovering over this would explain the tier system and what tier that video is.

As for working on the back end, it should know how it is deriving the dislikes and shouldn't be super hard, but then again i have no experience working on projects like this.

### Can you work on this?

- [ ] Yes
- [X] No

mattphelps-8451:
@KeinDahLahzey 
About Tier 2: There will always be the possibility for user extension dislikes on a video that has a large percentage of archived dislikes. Maybe this tier could set a threshold i.e. "At least 50% of dislikes on this video were archived" (and the other 50% or less are extension user-provided)

About Tier 4: Would you imagine dislikes not showing? Or an estimate of dislikes instead?

About Tier 4: Would

KeinDahLahzey:
@mattphelps-8451 

There might actually may need to be another tier in between the existing 3 and 4. This one would be "Dislikes were archived but not enough extension users to estimate further dislikes". Though having the percentage of dislikes from extension users and the archive would be handy.

As for dislikes showing on tier 4, I imagine that you would hover over and then it would tell you its estimate, but also that it may be wildly inaccurate since not enough data has been sampled. This would be to provide emphasis on the fact that it is inaccurate instead of people just looking and ignoring the accuracy.

sy-b:
similar to #114
but that tier idea is better that percentage

Anarios:
I will close this one because #221 is the same, but we can continue there. The basic idea is very similar.

PR #560, opened 2022-04-21
Implemented option to show like/dislike percentage instead of full like/dislike count

ErykDarnowski:
Issue #211 
I've had the same idea for some time, so after seeing @Nightcaat's issue I decided to work on this feature.

### DIFF
--- a/Extensions/combined/ryd.background.js
+++ b/Extensions/combined/ryd.background.js
@@ -277,6 +277,12 @@ function storageChangeHandler(changes, area) {
       changes.numberDisplayReformatLikes.newValue
     );
   }
+  if (changes.showTooltipPercentage !== undefined) {
+    handleShowTooltipPercentageChangeEvent(changes.showTooltipPercentage.newValue);
+  }
+  if (changes.numberDisplayReformatLikes !== undefined) {
+    handleNumberDisplayReformatLikesChangeEvent(changes.numberDisplayReformatLikes.newValue);
+  }
 }
 
 function handleDisableVoteSubmissionChangeEvent(value) {
@@ -291,6 +297,17 @@ function handleDisableVoteSubmissionChangeEvent(value) {
 function handleNumberDisplayFormatChangeEvent(value) {
   extConfig.numberDisplayFormat = value;
 }
+
+function handleShowTooltipPercentageChangeEvent(value) {
+  extConfig.showTooltipPercentage = value;
+};
+
+function handleTooltipPercentageModeChangeEvent(value) {
+  if (!value) {
+    value = "dash_like";
+  }
+  extConfig.tooltipPercentageMode = value;
+};
 
 function handleNumberDisplayRoundDownChangeEvent(value) {
   extConfig.numberDisplayRoundDown = value;
@@ -333,6 +350,8 @@ function initExtConfig() {
   initializeNumberDisplayFormat();
   initializeNumberDisplayRoundDown();
   initializeNumberDisplayReformatLikes();
+  initializeTooltipPercentage();
+  initializeTooltipPercentageMode();
 }
 
 function initializeDisableVoteSubmission() {
@@ -392,6 +411,26 @@ function initializeNumberDisplayFormat() {
       api.storage.sync.set({ numberDisplayFormat: "compactShort" });
     } else {
       extConfig.numberDisplayFormat = res.numberDisplayFormat;
+    }
+  });
+}
+
+function initializeTooltipPercentage() {
+  api.storage.sync.get(["showTooltipPercentage"], (res) => {
+    if (res.showTooltipPercentage === undefined) {
+      api.storage.sync.set({ showTooltipPercentage: false });
+    } else {
+      extConfig.showTooltipPercentage = res.showTooltipPercentage;
+    }
+  });
+}
+
+function initializeTooltipPercentageMode() {
+  api.storage.sync.get(["tooltipPercentageMode"], (res) => {
+    if (res.tooltipPercentageMode === undefined) {
+      api.storage.sync.set({ tooltipPercentageMode: "dash_like" });
+    } else {
+      extConfig.tooltipPercentageMode = res.tooltipPercentageMode;
     }
   });
 }


ErykDarnowski:
My bad, force-pushed to fix 2 things I missed.

Extensions/combined/popup.html
```diff
@@ -160,6 +160,34 @@ <h1 style="margin-bottom: 0.75rem" title="__MSG_extensionName__">
...
+  <br/>
+  <label class="switch" data-hover="Display percentage in like/dislike bar tooltip.">
+    <input type="checkbox" id="show_tooltip_percentage"/>
+    <span class="slider"/>
+    <span class="switchLabel">Percentage in like/dislike bar tooltip.</span>
+  </label>
+  <div class="custom-select">
+    <label for="tooltip_percentage_mode" data-hover="Use custom percentage display on hover.">Percent mode:</label>
+    <select name="tooltip_percentage_mode" id="tooltip_percentage_mode">
+      <option value="dash_like" id="tooltip_percentage_mode_dash_like">Dash like</option>
```
cyrildtm, line 172: Will it be easier for you and users if the <option> innerHTML is indeed the example: `190&nbsp;/&nbsp;10&nbsp;&nbsp;-&nbsp;&nbsp;5%`

cyrildtm:
Just want to point out this feature is already implemented in this extension: [Thumbnail Rating Bar for YouTube™](https://chrome.google.com/webstore/detail/thumbnail-rating-bar-for/cmlddjbnoehmihdmfhaacemlpgfbpoeb) I've been using it since the same time I got RYD. 

Anyway, good work.

cyrildtm:
BTW could you take a look at #540? Your feature would be also great for userscript.

ErykDarnowski:
Yeah, I saw your issue so I'm planning to do that ;). Also I think it's still nice to not have to install a couple extension for essentially features related to each other.

cyrildtm:
I'm not so worried. I use that extension mainly for batch examining. It happens to put a percentage in the ration bar.

Extensions/combined/popup.html
ErykDarnowski: Ohh, I didn't think about that... Let me try it out.

ErykDarnowski:
Implemented the feature ^ in UserScript.

ErykDarnowski:
I've made a typo in this commit's 9982082 msg, I meant issue #211 instead of #221

Extensions/UserScript/Return Youtube Dislike.user.js
```diff
@@ -29,6 +29,9 @@ let previousState = 3; //1=LIKED, 2=DISLIKED, 3=NEUTRAL
 let likesvalue = 0;
 let dislikesvalue = 0;
 
+// Temporary user config variable until better solution is found:
+var tooltipPercentageDisplayOption = "classic"; // classic ; dash_like ; dash_dislike ; both ; only_like ; only_dislike
```
cyrildtm, line 33: For future consideration, you may take a look at #540. We can mirror the data structure in the extension.

Extensions/UserScript/Return Youtube Dislike.user.js
```diff
@@ -204,6 +207,23 @@ function createRateBar(likes, dislikes) {
...
+  const tooltipPercentageDisplayModes = {
+    "classic": `${likes.toLocaleString()}&nbsp;/&nbsp;${dislikes.toLocaleString()}`,
+    "dash_like": `${likes.toLocaleString()}&nbsp;/&nbsp;${dislikes.toLocaleString()}&nbsp;&nbsp;-&nbsp;&nbsp;${likePercentage}%`,
+    "dash_dislike": `${likes.toLocaleString()}&nbsp;/&nbsp;${dislikes.toLocaleString()}&nbsp;&nbsp;-&nbsp;&nbsp;${dislikePercentage}%`,
+    "both": `${likePercentage}%&nbsp;/&nbsp;${dislikePercentage}%`,
+    "only_like": `${likePercentage}%`,
+    "only_dislike": `${dislikePercentage}%`
+  };
+
+  if (tooltipPercentageDisplayModes[tooltipPercentageDisplayOption] === undefined) {
```
cyrildtm, line 222: This won't be necessary for userscript, since everything is in this script :)

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: Sure, I'll probably work on this today 😉

ErykDarnowski: I'm wondering if it would be a good idea to add a boolean value for switching the percent display on/off like I've done it in the extension popup window. Although it could totally be done in just one var like it is at the moment.

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: Hmm... are you recommending a switch? Because either way (extension or UserScript) we still need to 'convert' the `tooltipPercentageDisplayOption` string from extConfig in to a format for the tooltip.

cyrildtm: No i meant everything will be defined unless user misuse! Undefined is used in extension for change events or first install/new feature update

Extensions/UserScript/Return Youtube Dislike.user.js
cyrildtm: What about adding a list option 'none'?

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: Ohh! I get what you mean. But I'm using `=== undefined` here to check if the string given by the user in `tooltipPercentageDisplayOption` from extConfig can be found as a key in the `tooltipPercentageDisplayModes` dict :). If the user inputs anything other than a correct option  `tooltipPercentageDisplayOption` is set to `"classic"` (the default value).

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: There is an option called `"classic"` already which works like `"none"` would :)

Extensions/UserScript/Return Youtube Dislike.user.js
cyrildtm: Yeah that's what I'm saying, you don't need it, because `tooltipPercentageDisplayOption` is set at the beginning, inside `extConfig` if you use the other pr. If it's undefined the switch statement should get into default.

cyrildtm: Actually i just read your code again. I would recommend against assigning strings for all options for every video, because you only need to see one. It's using a lot of computing power to process toLocaleString(), and the result will just get wasted.

Extensions/UserScript/Return Youtube Dislike.user.js
cyrildtm: The color thing has a separate switch so you can control two things differently, thumbs and bar. If i made the "none" option in the color theme then i can only turn everything on and off together.

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: Right, I missed that somehow, guess I should use a switch.

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: Yeah, yeah I get that. I was just wondering if it should replicate the extension popup, but there isn't really a good reason for it.

Extensions/UserScript/Return Youtube Dislike.user.js
ErykDarnowski: Fixed!

ErykDarnowski:
Hey @cyrildtm I'm trying to fix conflicts after your pull request #549 has been merged and I'm having an issue with a missing function called `localize` that should be in the utils.js file. What's up with that?

ErykDarnowski:
Actually, I found the function in your fork so no worries.

ErykDarnowski:
Actually, I found it already in your fork.

Blueberryy (approved):
Cool feature

ErykDarnowski:
It was @Nightcaat's idea I just implemented it, but thank you 😄

Issue #114, opened 2021-11-30
(Feature Request): Add info about dislikes and Trust level of info


sy-b:
### Extension or Userscript?

Both

### Request or suggest a new feature!

- [ ] Report Dislike information like "Last updated" & # 108, etc

- [ ] Give the dislikes a "Trust Score" depending on the information and reporting/counting method.

This will surely require us to come up with an equation for that rating.

### Ways to implement this!

Add an "ⓘ" (i in circle) button beside dislike counts. On hovering or clicking it should show info something like this.
(I don't know why adding small ⓘ  results in capital Ⓘ)
![Slide1](https://user-images.githubusercontent.com/94835959/144078326-0d6e217d-2aaf-4795-80ff-f4bb40354fd0.PNG)
(The rating can be in a range instead of percentage. Like 3/5)

### Can you work on this?

- [ ] Yes
- [ ] No

sy-b:
Question to People:
Any Ideas about how to rate the trust level/score?

FM1337:
What is the trust rating for exactly? 
- Is it for the video in question? 
  - is the video trustworthy? 
    - What makes it trustworthy? 
    - Does it even need to be rated to be trustworthy, what if it isn't a video that needs a trustworthy score? (Such as an animation video for fun)
- Is it for the creator?
   - Does the creator produce and publish content that requires a trustworthy score?
      - E.g. tutorials videos or videos featuring downloads
   - In the screenshot above, you have a "Creator Reported" status, could this be open to abuse if you were to allow users to report creators? (See below about users dislike bombing as this could also happen to lead to report bombing as well)
- Is it for the dislike themselves?
   - Are the dislikes reflective of the video or the creator?
   - Are the dislikes warranted?
      - Are actual users disliking the videos because they dislike the video or are they dislike bombing the video because of an outside factor?
         - Such as a controversy involving the creator or because someone decided to be funny and tell their fanbase to dislike someone's (or their own) video(s)
    - Are these actual users who are clicking dislike or is it a bot/script?
       - How would we tell if it was? User agent checking? Require an account? Use captchas? Check YouTube account APIs and use a scoring system to determine if this is a legitimate user or a bot?

tvelk:
[...]

PR #564, opened 2022-04-24
Add Dislike Information Panel

sy-b:


sy-b:
Note: Putting button inside YT Player -> idea from [SponsorBlock](https://github.com/ajayyy/SponsorBlock)

sy-b:
Please suggest a better UI or make one if you can.

cyrildtm:
Speaking of trust level, perhaps we can come up with an empirical model with some maths and programming. if some-- as many as possible-- video uploaders can share their studio data , we may examine the links between these variables: subscribers, views, likes dislikes, ratio. We assume the more extension users the more accurate, but by how much? This quantified concept can become trust level

sy-b:
> video uploaders can share their studio data

This starts the question of "**secure** sharing & **verification** of data".

> we may examine the links between these variables: subscribers, views, likes dislikes, ratio

While writing #114, **I _was_ assuming that creators would upload the data** on a regular basis and we'll use these variables
- last upload time & date (trust/accuracy level decreases with time)
- Data upload frequency
- Data Source & their percentage
    (Now mostly useless & very complicated to implement satisfactorily)
  Data sources to consider
    - YT API data before it was pulled down
    - Archived data by archiveteam.org
    - Creator (authorizing RYD to collect their YT data)
    - Extension user votes
- Extension users percentage for that video
- Video upload time (requires the variables you mentioned)
  Recently uploaded videos tend to change vote count rapidly. This depends on creator's audience size.


<br>

  After the idea of TLS modification method
- Upload method (App using API Key / Creator Studio)
  This is also mostly useless as the trust/accuracy level for that time must be set to 100%. But it's a bit easier to implement as we'll just have to consider time and **optionally** Upload frequency & extension user percentage as these factors wouldn't have much impact.
And if the creator's app is uploading it every second (YT API allows10k units/day) then there is no need of calculating accuracy level. It can be set to 100%. I find this unrealistic as the server load might be extreme but 1 upload per 5 minutes seems fine. And if possible, this per sec upload can be done for live streams.

---

This now seems unrealistic to me.

cyrildtm:
If we come up with an empirical model, we can just use much fewer existing video sentiments to calibrate it, so no need to constantly upload latest counts for every video. Using a model solves the eternal problem of sharing vs security.

Nightcaat:
Converting to draft as it is not ready yet.

I think the logo in the player could be a tad smaller to match everything else better

sy-b:
> Converting to draft as it is not ready yet.

👍


> not ready yet.

Depends on @Anarios. I am waiting for the changes.

xHeaveny (commented):
I would use Socal Blade for this ngl

sy-b:
👍

MulverineX:
While this more complex UI is useful, a simplified version on hover of the Like/Dislike/Ratio area would be appreciated also, perhaps something like this:
Group 113 (2)
Group 113 (3)
Group 113 (4)

MulverineX:
In fact, perhaps instead of embedding that giant panel for detailed data, that can go in the extension popup?

cyrildtm:
> go in the extension popup

That would require message exchange, which is a bit complicated, but it's doable, if desired. 
That said, I don't think I will want to spend more than a split second to look into the dislike data.

sy-b:
Pros & Cons of this panel

- Pros
  1. People had requested variety of things. (I can't even recall 80% of them.) _**Maybe**_ this _can_ act as a one size fits all approach.
  2. Will work on devices where you can't hover over the button. (It can't rn)

- Cons
  1. Bulky UI
  2. The current way of implementation isn't good for most of the use cases.

<br>


I am inclined to reimplement this the way SponsorBlock has <ins><b>if</b></ins> we extpand its purpose. 

from: https://github.com/ajayyy/SponsorBlock/blob/0c216792cf1737e13e358411902f4b77dce1725e/src/content.ts#L2095-L2112
```JS
function sendRequestToCustomServer(type, fullAddress, callback) {
    const xmlhttp = new XMLHttpRequest();

    xmlhttp.open(type, fullAddress, true);

    if (callback != undefined) {
        xmlhttp.onreadystatechange = function () {
            callback(xmlhttp, false);
        };

        xmlhttp.onerror = function() {
            callback(xmlhttp, true);
        };
    }

    //submit this request
    xmlhttp.send();
}
```
Also refer: https://github.com/ajayyy/SponsorBlock/blob/0c216792cf1737e13e358411902f4b77dce1725e/src/content.ts#L1666-L1733

This way we can have the same pages shown in popup & panel. Also, having these pages in a separate folder will be better.

sy-b:
> a simplified version on hover of the Like/Dislike/Ratio area would be appreciated

Yeah, but this is inaccessible to mobile users.

<br>

> perhaps instead of embedding that giant panel for detailed data, that can go in the extension popup?

Judging by the amount of people asking "How to get to the extension setting menu for colors", I can very well say that many people can easily find chrome://extension/whatever-extension-id than a settings button in the extension popup.
I am still inclined to that idea (panel(s) in popup).

<br>

>  I don't think I will want to spend more than a split second to look into the dislike data.

I agree.

MulverineX:
> Yeah, but this is inaccessible to mobile users.
Is this really a use case we should consider? Vanced is by no means dead and most others use 3rd party apps that add functionality like RYD. YouTube Web Mobile is trash

cyrildtm:
> YouTube Web Mobile is trash

So true! I can't even watch a video full screen. Not to mention potato touch control. I never use mobile web version, avoid like plague

sy-b:
> Is this really a use case we should consider?

Initial contributors did consider supporting it & most of the extension features are supported on mobile.  
But yeah, idk why people will use mobile UI. ¯\\_(ツ)_/¯ .

MulverineX:
Maybe there's a way to detect long press?

Sidchou:
would be great if hovering the bar can shows more info like 
https://github.com/Anarios/return-youtube-dislike/pull/564#issuecomment-1108864779
we just need number for user %

PR #572, opened 2022-04-25
Opening `changelog_3.0.html` on extension update

avi12:
Currently, the HTML page opens every time the extension updates _or_ the browser updates
This PR is for changing this behavior so only when the extension updates, the HTML page will open up


cyrildtm:
Hi @avi12 I understand you might be angry about the annoying popups, but the main discussion is in #537, with similar implementations to yours. Please take a look.

Anarios:
Fixed in #537

PR #574, opened 2022-04-27
Fix userLocales is undefined

hyperstown:
With tempermonkey on mobile firefox user.js script returned error:
`userLocales is undefined` which caused dislikes to stay always at 0.

I noticed that variable is not accessible from `getNumberFormatter` function.
I'm not sure if this is the best approach but I've passed  `userLocales` as parameter. 
I've also set `userLocales` to `document.documentElement.lang` if array search fails.


Anarios:
Thanks

PR #591, opened 2022-05-07
Integrate Readme translation into Spanish

Alejandrogayol:
This will show a link to display the Readme file in Spanish. Another people can made more translations to different languajes and add the links in the same section (“Read this in another languaje”).
Translation file is “READMEes.md”.


Anarios:
Thanks <3

PR #610, opened 2022-05-14
Change languajes to languages

SrIzan10:
little "extension" on the #591 PR.

Changed languajes to languages


SrIzan10:
yep, sorry

PR #627, opened 2022-05-27
Shorter/relative paths in Markdown files

Sky-NiniKo:
### I think it makes more sense to use a relative path:
**\+** It shorter. (me)
**\+** It allows to refer to the corresponding file in a fork and not to the original file. (me)
**\+** Relative paths will function even if I move a directory tree into another (@​sy-b https://github.com/Anarios/return-youtube-dislike/pull/627#issuecomment-1140177384)

**\-** If you download only part of the files to your local machine: the links are broken. (@sy-b https://github.com/Anarios/return-youtube-dislike/pull/627#issuecomment-1139306332)
**\-** More ambiguity and can lead to confusion (@cyrildtm https://github.com/Anarios/return-youtube-dislike/pull/627#issuecomment-1139806538)
**\-** Complete path can be understood at a glance (@​sy-b https://github.com/Anarios/return-youtube-dislike/pull/627#issuecomment-1140177384)

### A vote ?
I don't know who has the last word on controversial PRs but I propose to vote either directly in this issue by reacting to this message with a like or a dislike or to open a [poll](https://github.com/Anarios/return-youtube-dislike/discussions/categories/polls).

### DIFF
--- a/Docs/readme.md
+++ b/Docs/readme.md
@@ -13,7 +13,7 @@
 - [Troubleshooting](https://github.com/Anarios/return-youtube-dislike/wiki/Troubleshooting-Guide)
 <!-- - [FAQ](FAQ.md)
 - [When & How to Report Bugs](Guide__Bug_Reporting.md)
-- [Contributing](/CONTRIBUTING.md) -->
+- [Contributing](../CONTRIBUTING.md) -->
 <!-- - [How to update wiki](/) -->
 
 <br>


sy-b:
Let's say [README.md](https://github.com/Anarios/return-youtube-dislike/blob/main/README.md) is on my local machine and I want to read [FAQ.md](https://github.com/Anarios/return-youtube-dislike/blob/main/Guides/FAQ.md). If I encounter relative path(s), I'll definitely have to manually hunt for the file online. Of course, the relative paths make more sense when I have the whole repo but its a bit troublesome when I don't or I have an outdated one. This is the only downside I know. For rest of the cases, its good.
For this particular case its 👍.

sy-b:
> I don't know who has the last word on controversial PRs

@‎anarios

sy-b:
Some more advantages:

- Relative Paths
  - Relative paths will function even if I move a directory tree into another
- Absolute Paths
  - Complete path can be understood at a glance.

sy-b:
For this particular case, it wouldn't a problem if we used relative path for [README.md](https://github.com/Anarios/return-youtube-dislike/blob/main/README.md) **only**. But I don't have a strong opinion yet.

sy-b:
In some cases `../something` can be dangerous.

I usually prefer to use `/something`.
The `../` in `../something` refers the 1st folder in the current the directory. And during directory shuffling, this can be a headache. **Use `../` with caution**.

PR #640, opened 2022-06-07
Add formula used for dislike calculation

sy-b:
### **5. How is the dislike count calculated?**

RYD uses the votes from it's users to extrapolate the dislike count.

- If the video was uploaded after the API was shut down:

  $$ \textup{RYD Dislike Count} = \left( \frac{\textup{RYD Users Like Count}}{\textup{RYD Users Dislike Count}} \right) \times \textup{Public Like Count} $$

- If the RYD database somehow had the actual like and dislike count (provided by the uploader or from the archive), the dislike count will be calculated based on both - the users' votes and the archived value. The archived value will have less influence on the final count as it ages.

---

This in video form

[IReturn YouTube Dislike Explained](https://www.youtube.com/watch?v=GSmmtv-0yYQ)

---

### DIFF
--- a/Docs/FAQ.md
+++ b/Docs/FAQ.md
@@ -25,7 +25,7 @@ RYD uses the votes from it's users to extrapolate the dislike count.
 
 - If the video was uploaded after the API was shut down:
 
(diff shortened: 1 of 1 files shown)


sy-b:
Thanks

PR #642, opened 2022-06-09
Add ratio preview to videos

aaronkerckhoff:
I added the ability to see the like/dislike ratio next to every video:
ffc260b8

### DIFF
--- a/Extensions/combined/ryd.content-script.js
+++ b/Extensions/combined/ryd.content-script.js
@@ -40,12 +40,18 @@ import {
   addLikeDislikeEventListener,
   storageChangeHandler,
 } from "./src/events";
+import {
+  showRatioPreviews,
+  resetVideoList,
+} from "./src/preview";
 
 initExtConfig();
 
 let jsInitChecktimer = null;
 
 function setEventListeners(evt) {
+  // Reset ratio preview video list on tab change 
+  resetVideoList();
   function checkForJS_Finish() {
     if (isShorts() || (getButtons()?.offsetParent && isVideoLoaded())) {
       addLikeDislikeEventListener();
@@ -66,3 +72,7 @@ document.addEventListener("yt-navigate-finish", function (event) {
   window.returnDislikeButtonlistenersSet = false;
   setEventListeners();
 });
+
+setInterval(function () {
+  showRatioPreviews();
+}, 3000);
--- a/Extensions/combined/src/preview.js
+++ b/Extensions/combined/src/preview.js
@@ -1,0 +1,44 @@
+import { getVideoId } from "./utils";
+import { getApiData } from "./state";
+
+let videos = [];
+
+async function showRatioPreviews() {
+  // Select all videos
+  let currentVideos = [].slice.call(document.getElementsByTagName("ytd-rich-grid-media"));
+  // Video recommendations while watching a video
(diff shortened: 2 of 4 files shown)


aaronkerckhoff:
> Just to let you know, there is a different extension that has this feature: [Thumbnail Rating Bar for YouTube™ ](https://chrome.google.com/webstore/detail/thumbnail-rating-bar-for/cmlddjbnoehmihdmfhaacemlpgfbpoeb)
> 
> In [the above extension](https://github.com/elliotwaite/thumbnail-rating-bar-for-youtube) they have a delay function to slow down the for-loop for RYD API server queries, and a local cache to temporarily store query results for up to ten minutes. They acknowledged the limits by implementing this feature so individual users don't get banned by RYD API server.
> 
> It's nice to see you are implementing this function here, but would you consider including the delay and cache as well? That may prevent massive server requests, which is basically 100x traffic amplification.
> 
> cache: https://github.com/elliotwaite/thumbnail-rating-bar-for-youtube/blob/master/extension/background.js delay: https://github.com/elliotwaite/thumbnail-rating-bar-for-youtube/blob/1ba786dcf6620310377d194b6b71919fbcf41d30/extension/content-script.js#L2

Thank you very much for that information. I also thought about that problem and thought that it would be a good idea to allow getting data for multiple videos at once. For example, I send a list with all videos on my start page [Sjfdyj5s, ...] and then get a list with all the data at once by the API. But as far as I know, the API isn't open source.

I'm not here for the weekend, but I'll work on the delay and caching, as soon, as I'm back and I really appreciate the links to the other extension.

sy-b:
Another idea from discord (by Ezed)- 
Most of the time users are going to ignore that extra info. Hence, it isn't necessary display it on all videos.
A better way can be to display the ratio on hover.

aaronkerckhoff:
> Another idea from discord (by Ezed)-
> 
> Most of the time users are going to ignore that extra info. Hence, it isn't necessary display it on all videos. A better way can be to display the ratio on hover.
> 
> We should probably add some delay to that too.

Thanks for this suggestion. This is a good idea. Maybe the ability to switch between the modes would be good idea? (Off, Only on hover, Always). With the "always-mode" using the delay and caching.

aaronkerckhoff:
> Maybe make a switch for the continuous-fire mode (with delay) and semi-auto mode (less or no delay).

Do you mean with "continuous-fire-mode" that the ratio is always displayed and with "semi-auto-mode" that it is only displayed on hover but then without delay?

aaronkerckhoff:
Alright, thank you. I will work on the different modes as well as the caching/delaying as soon as I can.

sy-b:
Great 👍

aaronkerckhoff:
Thank you very much for all your suggestions! I'm currently working on the implementation.

aaronkerckhoff:
Thank you very much for all of your suggestions and the time you spend on them. I fixed everything you mentioned.

> What's your plan with variable `videos`?
In an interval of 3 seconds, all videos on the page are gathered. Every video then gets the addition of the ratio display. As soon as it displays the ratio, it gets added to the list `videos`. This is to prevent every video from being updated every 3 seconds, but only new videos (for example when scrolling down).

aaronkerckhoff:
Sorry that I didn't update this PR sooner, I was really busy in the last few months. I now added the ability to toggle the different modes: always, hover, never. Everything works fine so far, if there are still some bugs, let me know. Thanks to all that helped me and contributed to this PR.

PR #673, opened 2022-07-10
Show "Dislike" instead of "disabled by owner"

JustOptimize:
The idea is to show the "Dislike" text when the Likes/Dislikes are disabled by owner to maintain the vanilla & minimal look of youtube


Anarios:
Closing due to reasons given by @cyrildtm

PR #708, opened 2022-08-01
Feature/add unit testing framework and start writing unit tests (acceptance tests)

leonbubova:
Hey everyone,

I had a look at the codebase again and there are some opportunities to refactor thinks that were build back in the days, but before we do that we should have tests in place to make sure we don't introduce bugs while refactoring.

That's why I added jest for automated unit and integration testing to the project. I didn't write tests for everything yet, but started with the `utils.js` since they were easy to take apart.

Two things I made sure to do:

- I tried to not touch the code that I was testing (other than 1 exception that was necessary -> see changes)
- I am basically writing acceptance tests only for now (Which means I am strictly documenting the current behaviour of the code (There is one example where I am testing behaviour that I would say is buggy currently -> see changes)

The reason behind that is, that it is easier to strictly cover the codebase with tests first and have them as a contract for existing code and later go ahead and change the codebase with the tests as security net.

I used jest as the testing framework because it's in my opinion the most widely used and we can expand it to integration tests as well later on as well (which we should do to minimize manual testing needed when releasing changes in the future)

Let me know what you think of the changes, I'll try and keep my changes conflict free until it get's merged (or declined) 

That's the current code coverage report:

image


Extensions/combined/src/state.js
```diff
@@ -46,42 +46,46 @@ function isNewDesign() {
...
-        mutation.target.id === "button"
-      ) {
-        // cLog('Short thumb button status changed');
-        if (mutation.target.getAttribute("aria-pressed") === "true") {
-          mutation.target.style.color =
-            mutation.target.parentElement.parentElement.id === "like-button"
-              ? getColorFromTheme(true)
-              : getColorFromTheme(false);
-        } else {
-          mutation.target.style.color = "unset";
```
leonbubova, line 76: This was the one place where I had to change existing code, to make testing possible.

The problem was, that this code was always executed whenever `state.js` was imported somewhere. But because in there are variables that were not existing in the test environment, there were errors all around the place.

I moved to code to it's own function `initMutationObserver()` and exported it, to be used in the one place where it should be called. That's what we should aim for when using modules in the future. Loose code otherwise will lead to problems.

Extensions/combined/src/utils.js
```diff
@@ -155,7 +155,9 @@ function getColorFromTheme(voteIsLike) {
 }
 
 export {
+  roundDown,
   numberFormat,
+  getNumberFormatter,
```
leonbubova, line 160: I exported those as well to make them testable

Extensions/combined/src/utils.spec.js
```diff
@@ -0,0 +1,341 @@
...
+
+  it("should return a correct formatter when using firefox", () => {
+    Object.defineProperty(document.documentElement, "lang", {
+      value: null,
+      configurable: true,
+    });
+    expect(getNumberFormatter()).toMatchSnapshot();
+  });
+
+  it("should return a correct formatter when using the URL locale", () => {
```
leonbubova, line 103: Extra info: I am planning to write the acceptance tests in this format as mentioned in the summary already, because it makes it easier to cover the codebase with tests. If I assume there is a bug in the code, I write the test so that they are green with the current code, but also add a section what the correct test code should be to trigger the buggy behaviour.

These can be addressed later on then, when code coverage is complete. By doing this approach we can minimize introducing new bugs to the codebase by refactoring.

jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  resetMocks: true,
+  collectCoverage: false,
+  collectCoverageFrom: [
+    "Extensions/combined/src/**/*.js",
+    "Extensions/combined/*.js",
+  ],
```
leonbubova, line 7: Disabled code coverage collection for now for performance reasons, but it can be enabled by changing `collectCoverage` to `true`

Extensions/combined/src/utils.js
leonbubova: What exactly do you mean by 
> does that mean it's better to do it in the source code?

But one thing to add: Typically if something is not exported, it means it's an implementation detail and should not be tested separately, but rather by testing the public/exported functions that use them. But in this case I decided to make them testable since the exported function is a rather big unit of code. But i someone disagrees we can always undo that and remove the tests for `roundDown` and `getNumberFormatter` again.

leonbubova: Ahh I see. Good question. There is not a general consensus on that question to be honest. Some say "There is a reason that a function is private/not exported", others say "it's better to have all code under test". Having everything under test can lead to having to rewrite the unit tests more often when you refactor, but for the state the codebase is in right now, I would lean to let's have everything testable. Because generally I would say the functions are doing quite a lot of things and once we have everything covered in tests I would love to break specific parts up in more atomic functions. Those could then be private/not exported/untested.

Anarios:
Amazing work, this was in my plans for a looong time.

leonbubova:
@Anarios (or anyone else who knows): I was gone for some time before opening this PR so I am not sure what the current approval/review workflow is. I see there are 6 reviews required. Do I have to actively request those, or what's the best process?

Anarios:
@leonbubova, sorry, I've missed your question, and generally forgot about this PR. 

I've resolved merge conflicts and merging this now (well, better late than never)

PR #759, opened 2022-09-18
QFIX hidden dislike count, extension & userjs

### LINKED ISSUE #757: YouTube dislike numbers not showing up
### Browser

Mozilla Firefox

### Browser Version

104.0.2

### Extension or Userscript?

Extension

v3.0.0.5

https://www.youtube.com/watch?v=DWWK05t98os

### What happened?

The dislike count is not showing whatsoever.

Refreshing the page works. This is happening on both Firefox and Chrome.

Yes

hyperstown:
Fixes #757

In some cases dislike count was hidden from the user. YT added attribute `is-empty` that had CSS property `display: none`
image
image

I've added `getDislikeTextContainer()?.removeAttribute('is-empty');` to `setDislikes()`

Other possible solutions would be to set css property.

Fix tested with both extension and userscript

### DIFF
--- a/Extensions/UserScript/Return Youtube Dislike.user.js
+++ b/Extensions/UserScript/Return Youtube Dislike.user.js
@@ -219,6 +219,7 @@ function setDislikes(dislikesCount) {
     mobileDislikes = dislikesCount;
     return;
   }
+  getDislikeTextContainer()?.removeAttribute('is-empty');
   getDislikeTextContainer().innerText = dislikesCount;
 }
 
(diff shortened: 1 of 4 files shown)


HAYD323:
This seems to be happening for both likes and dislikes FYI

hyperstown:
> This seems to be happening for both likes and dislikes FYI

I could not reproduce the situation where both like and dislike disappears and although I could assume that current solution would also work for likes there's no way for me to test this. 
If someone would provide me with some analysis of the code in such a state I could push the fix.
Let's see what maintainers will say.

Anarios (approved):

Anarios:
Thank you

Issue #752, opened 2022-09-13
Buggy on Latest YouTube UI


futureflash01:
### Browser

Google Chrome

### Browser Version

105.0.5195.102

### Extension or Userscript?

Extension

### Extension/Userscript Version

3.0.0.5

### Video link where you see the problem

https://www.youtube.com/watch?v=Y3fi9pc81NY

### What happened?

The dislike count still shows, however there's still the "Dislike" text showing next to the number.

In other words, Return YouTube Dislike is not fully supported for the latest YouTube UI update.

### How to reproduce/recreate?

1. Open YouTube
2. Click on any video
3. Bug shows instantly

![image](https://user-images.githubusercontent.com/50788891/189975576-bff50aff-117e-4642-ae94-e5ab76189689.png)

### Will you be available for follow-up questions to help developers diagnose & fix the issue?

Yes

unbeatable-101:
This looks the same as #742

vizthex123:
Ah, I was gonna come post about this issue too. Some tabs keep the normal UI, and some do this weird bug.

And oh god, is this the new YouTube UI? Please tell me there's a way to change it back.

TukangM:
![image](https://user-images.githubusercontent.com/91467886/191740228-f80926f7-ee14-45ad-bfcc-94b1d6ae10c1.png)
 shorts video

PickleNik:
seems to be fixed in latest version

PR #761, opened 2022-09-19
FIX: Dislike button displaying "Dislike" text next to count

### LINKED ISSUE #742: Dislike button displaying "Dislike" text next to count.
### Browser

Chrome

### Browser Version

104.0.5112.102 (Official Build) (64-bit) (cohort: Stable) 

### Extension or Userscript?

Extension

3.0.0.5

https://www.youtube.com/watch?v=PzOLDt8fLT8 (but this can be reproduced in every video)

### What happened?

Every video's Like/dislike bar shows the format "Dislike[count]" instead of just the count. Please see the attached screenshot. 

chrome_iWAUlFcBOX

I have also formatted my PC yesterday, so I can confirm it can be reproduced both before and after the clean up.

Just need to open any video on Youtube.

Yes

hyperstown:
Fixes #742, #752

New design now has replaced `id='text'` with an attribute `text`. 

`getDislikeTextContainer()` could not find dislike container since it was looking for ID or specific tag name add added additional span with dislike number.

I've added additional `querySelector` that checks for that attribute as well.

UserScript has this already implemented in my previous PR.

### DIFF
--- a/Extensions/combined/src/buttons.js
+++ b/Extensions/combined/src/buttons.js
@@ -56,7 +56,8 @@ function getDislikeButton() {
 function getDislikeTextContainer() {
   let result =
     getDislikeButton().querySelector("#text") ??
-    getDislikeButton().getElementsByTagName("yt-formatted-string")[0];
+    getDislikeButton().getElementsByTagName("yt-formatted-string")[0] ??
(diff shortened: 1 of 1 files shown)


Anarios (approved):

Anarios:
Thanks

Issue #733, opened 2022-08-25
Offset Like/Dislike ratio bar with updated youtube UI


chloe-burbank:
### Browser

Firefox

### Browser Version

Firefox 103.0.2

### Extension or Userscript?

Extension

### Extension/Userscript Version

Version v3.0.0.5

### Video link where you see the problem

https://www.youtube.com/watch?v=SrcupdKGvBs

### What happened?

Like/dislike ratio bar is offset to the left of the like/dislike buttons instead of right below them

![image](https://user-images.githubusercontent.com/94586008/186773937-0bac4d73-1c23-453a-92bf-21292a335fdd.png)

### How to reproduce/recreate?

1. Go to youtube.com
2. Click on any video

### Will you be available for follow-up questions to help developers diagnose & fix the issue?

Yes

taikonaut001:
![image](https://user-images.githubusercontent.com/60836715/186788306-74f4c9d7-4cd0-45b6-88aa-0c359d1267ad.png)

cjfaquino:
Got the same results with Chrome and Edge. It seems like you have to be logged in to see the new UI too, otherwise it functions like normal.

slvr2314:
I have the same issue on both Firefox and Brave. But like @cjfaquino said, if I'm logged out of Youtube, the alignment works just fine.

ErykDarnowski:
Hey, I've actually only managed to reproduce this issue on Firefox, and funnily enough, in my case it happened the other way around (meaning, when I was logged out instead of logged in).

Anyway, I've found out that it's caused by the `#actions-inner` element stretching:

https://user-images.githubusercontent.com/81530705/194659895-dfdc386d-e0f4-45c1-8904-7acd6dd76d35.mp4

This can be fixed by changing its `id` from `actions-inner` to `menu-container` or simply removing the `ytd-watch-metadata` from its `classList`:

![image](https://user-images.githubusercontent.com/81530705/194669513-5d945577-3e81-4940-bd12-9be883db34da.png)

```javascript
document.getElementById('actions-inner').classList.remove('ytd-watch-metadata');
```

---

**I'll probably work on implementing this fix tomorrow.**

ErykDarnowski:
Ok, after further testing (which also included the new rounded UI) I've improved my solution to something along the lines of this:

```javascript
if (isNewDesign()) {
  document.getElementById("actions-inner").style.width = "revert";
					
  // Checking if it's the new (rounded) UI
[...]

PR #771, opened 2022-10-08
Fix like/dislike ratio bar offset

ErykDarnowski:
Fixes [#733](https://github.com/Anarios/return-youtube-dislike/issues/733)

I also wanted to implement this fix in **mobile** and **UserScript** but after seeing how far behind they were and that some PRs were in progress to change that, I've decided to hold off on that for now.


Anarios:
Thanks a lot

PR #803, opened 2022-11-01
GitHub Issues: Ensuring a bug report is not a duplicate on the first stage

Moondarker:
Wanted to add it at the bottom at first, then thought about potential time waste for those who file an issue and it ends up being a duplicate - it's better to keep it at the top


YassineGherbi:
Pretty crucial imo. Great improvement!

YassineGherbi (approved):

PR #808, opened 2022-11-03
Dislike text margin

BaraaHalabiTr:
I added a littler margin to the text to better accommodate youtube's new theme.


Anarios:
Would break other UI styles (the margin would be too big)

Arora-Sir (commented):
#827 has solved this issue with 6px margin at left, and without breaking the other UI. You can approve that and do the same here I guess.

Anarios:
Thanks

PR #816, opened 2022-11-08
Fix "API down" error message

hyperstown:
Hi, I fixed invalid query selector in `displayError` function. Now it shows error message.

image


Anarios (approved):

Anarios:
Thank you

PR #829, opened 2022-11-22
Fix disappearing rate bar

hyperstown:
Hi, I've noticed that ratio bar disappears when youtube uses it's push history (doesn't load whole website from zero).
It turns out that it doesn't actually disappears but its being pushed outside visible canvas. The way youtube updates DOM, ratio bar is somewhere inside hidden elements.

Currently rateBar function just checks if rateBar `div` exists and updates it if it does.
But if rateBar is outside visible canvas update is pointless.

```js 
let rateBar = document.getElementById("ryd-bar-container");
```
image
(Firefox grays out invisible elements in inspector)
image

I've added additional check that ensures rateBar is visible. If not. Removes old bar and adds a new one.
```js
if(rateBar && !isInViewport(rateBar)){
      rateBar.remove();
      rateBar = null;
    }
```
Second issue was that rateBar position selector found invisible buttons container before actual one.
 `getElementById()` assumes that ID is unique per page so it returns first element that matches. Unfortunately  youtube has multiples IDs with the same name. Just one of them is visible though.

```js
document.getElementById(
            isNewDesign() ? "top-level-buttons-computed" : "menu-container"
 ) || document.querySelector("ytm-slim-video-action-bar-renderer")
 ```

image

I could iterate every element and check if is visible but in new design we already have good working selector that returns `top-level-buttons-computed` id. `getButtons`

I've added this selector:
```js
let actions = isNewDesign() && getButtons().id === "top-level-buttons-computed" 
          ? getButtons() : document.getElementById("menu-container");
actions || document.querySelector("ytm-slim-video-action-bar-renderer")
```
Perhaps we could remove `&& getButtons().id === "top-level-buttons-computed" ` but the way youtube handles everything I don't know if that's a good idea.

Result:

image

### DIFF
--- a/Extensions/combined/src/bar.js
+++ b/Extensions/combined/src/bar.js
@@ -14,7 +14,6 @@ function createRateBar(likes, dislikes) {
   if (!isLikesDisabled()) {
     // sometimes rate bar is hidden
     if(rateBar && !isInViewport(rateBar)){
-      cLog('create rateBar remove old')
       rateBar.remove();
       rateBar = null;
     }
--- a/Extensions/combined/src/bar.js
+++ b/Extensions/combined/src/bar.js
@@ -10,11 +10,9 @@ import {
 import { cLog, getColorFromTheme } from "./utils";
 
 function createRateBar(likes, dislikes) {
-  let ratebar = document.getElementById("ryd-bar-container");
-  if (ratebar) {
-    ratebar.parentNode.removeChild(ratebar);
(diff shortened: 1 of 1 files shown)


Rizxcviii:
Fixes #837 but I've also made another PR attacking this issue another way (#838)

Anarios:
Amazing work, sorry for taking long to merge, will be released soon

PR #838, opened 2022-11-26
fix: ratio bar not rendered on initial navigation towards video

### LINKED ISSUE #837: Ratio bar only loads on refresh of browser tab
### Have you tried to find similar open issues?

- [X] Yes, this issue is not a duplicate

### Browser

Opera GX and Google Chrome

### Browser Version

Chrome/106.0.0.0 and Chrome/107.0.0.0 respectively

### Extension or Userscript?

Extension

3.0.0.6

All of them

### What happened?

It appears on the first load of a browser TAB (that's right, every tab, not just a window) the ratio bar will not load. Only on a refresh of the browser tab, will the ratio bar appear. Not even switching to another video will work. Only appears from a browser refresh, from what I've gathered.

I'm currently using the main branch as well, so all updates must be present on my version.

Load a new browser tab, and you will notice the ratio bar missing. Refresh the browser tab for the ratio bar to appear.

Yes

Rizxcviii:
The wrong element was being targeted. Apparently (maybe in a recent YouTube update), there was an additional custom element `ytd-browse` that was added; containing the home page content. This element was semantically placed **ABOVE** the `ytd-watch-flexy` custom element that contained `/watch` content. This was probably to ensure that the entire page (sidebars, etc.) wouldn't re-render on navigation changes. For now, this specifically targets the correct parent element.

Fixes #837

### DIFF
--- a/Extensions/combined/src/bar.js
+++ b/Extensions/combined/src/bar.js
@@ -56,14 +56,21 @@ function createRateBar(likes, dislikes) {
           colorDislikeStyle = "; background-color: " + getColorFromTheme(false);
         }
 
-        (
-          document.getElementById(
-            isNewDesign() ? "top-level-buttons-computed" : "menu-container"
-          ) || document.querySelector("ytm-slim-video-action-bar-renderer")
-        ).insertAdjacentHTML(
+        const parent = document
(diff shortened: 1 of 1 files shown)


Anarios:
I went with the solution from #829, sorry.

PR #863, opened 2023-01-18
Added a german version of the README

Wuffli:
+ Added a german version of the README
+ Added links to german version on the other versions
+ Added links to the extention storepages [Only in german version, maybe unwanted?]

Proofread it once myself, there are most likely some errors left.


Anarios:
Danke

PR #865, opened 2023-01-25
Add Korean translation

liberaldev:
Added Korean translations for extensions and website. I've also tested it in action and it works well in my environment.

Also added .nvmrc (https://github.com/nvm-sh/nvm#nvmrc) as a bonus, CONTRIBUTING.md (https://github.com/Anarios/return-youtube-dislike/blob/main/CONTRIBUTING.md) was added as the recommended 12.18.4 version.


Anarios:
Thanks!

PR #874, opened 2023-02-16
Show dislikes again on new YT UI (01) & fix workflow faliure

sy-b:
# 1. Show dislikes again on one of the new YT UI layout

## After merging this PR
New

## Current main branch
Old

 ---

# 2. Fix failing weblint workflow

```
13  /home/runner/work/return-youtube-dislike/return-youtube-dislike/Website/nuxt.config.js
14     8:8   error  'de' is defined but never used  no-unused-vars
15    54:46  error  'uk' is not defined             no-undef
```

---

- The title has `(01)` as I expect similar PRs in future.

PR #886, opened 2023-03-02
Show dislikes again on new YT UI (01) (2)

sy-b:
fixed https://github.com/Anarios/return-youtube-dislike/pull/874


Anarios (approved):
Thanks

PR #888, opened 2023-03-05
fix get count dislikes on mobile version of youtube

TalosDx:
The pull request fixes getting dislikes on the mobile version of the site.
It fixes it in two ways.
1. Checking the existence of the dislikes button before adding a listener to increase the number of dislikes.
2. An alternative solution, in case the error still occurred. (Possibly requires a fix, because it is performed without delay)

Problem founded in compiled code:
изображение

```js
buttons.children[0].children[0].addEventListener("click", likeClicked);
buttons.children[0].children[1].addEventListener("click", dislikeClicked);
buttons.children[0].children[0].addEventListener("touchstart", likeClicked);
buttons.children[0].children[1].addEventListener("touchstart", dislikeClicked);
```

buttons.children[0].children[1] is undefined

Why getDislikesButtons() compiled to buttons.children[0].children[1] i don't know.

### DIFF
--- a/Extensions/combined/ryd.content-script.js
+++ b/Extensions/combined/ryd.content-script.js
@@ -44,16 +44,24 @@ import {
 initExtConfig();
 
 let jsInitChecktimer = null;
+let isSetInitialStateDone = false;
 
 function setEventListeners(evt) {
   function checkForJS_Finish() {
     if (isShorts() || (getButtons()?.offsetParent && isVideoLoaded())) {
       addLikeDislikeEventListener();
       setInitialState();
+      isSetInitialStateDone = true;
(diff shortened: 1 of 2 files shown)


TalosDx:
It is also worth looking into that the code compiles. It is possible that the problem is in the compilation.

Anarios:
Thanks a lot

TalosDx:
After being accepted into the main branch, it is also necessary to collect it into a single file (dir UserScript), I tried to do it locally, but for some reason it was collected in several different files.

ranazee:
[Extensions/combined/ryd.content-script.js](Extensions/combined/ryd.content-script.js)

ranazee:
Solve this iss8

PR #487, opened 2022-02-20
Added Safari Web Extension (iOS, iPadOS, macOS)

1998code:
## Tested on:
- [x] iOS / iPadOS 15.3
- [x] macOS 12.3

# Get Started with Safari web extension

| iOS / iPadOS | macOS |
| ---- | -----|
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-02-20 at 16 10 25 | CleanShot 2022-02-20 at 16 19 18@2x |

# How to enable the extension
Safari > Preferences... > Extensions

CleanShot 2022-02-20 at 16 12 30@2x

PR #443, opened 2022-01-15
Add safari extension

onodera-punpun:
I know there is #133, but that PR seems a bit stale. This one is up to date with master, and makes use of webpack to build the safari extension.

Note that I have only tested this on MacOS, and not iOS.

Closes #127.

PR #133, opened 2021-12-01
Add Safari extension

rdev:
Adds Safari extension
Converted from Firefox using Xcode web extension converter with no additional changes besides app icon adjustment 😄 

Tested and confirmed working on macOS Monterey, not tested (yet) on iOS/iPadOS. Might require tweaks there since iOS does not support persistent background pages and conversion threw up a warning

### Screenshots

Screen Shot 2021-12-01 at 18 26 23

Screen Shot 2021-12-01 at 18 26 26

### App Store
I can publish it to the App Store with permission from maintainers if needed 

re: #127

PR #896, opened 2023-03-30
Add Safari extension build target

suhailskhan:
This is an alternative implementation to those in #133, #443, and #487.

Similarly to #443, this adds a Safari extension as a webpack plugin pattern. It then generates an Xcode project for the macOS and iOS apps using a build script in `package.json`.

Building the Safari extension now requires only one short command. Running `npm run build:safari` will run the standard webpack build steps, and then invokes the Xcode Safari extension generator to create an Xcode project. All files generated by doing this will be found under `Extensions/combined/dist`. At this point, the Xcode project opens automatically and one can compile and export copies of the Mac and iOS Safari extensions.

What I did differently from the other PRs is I treated the Xcode project as a build artifact, rather than treat it as part of the codebase for the extension. Not only does it keep the repo’s directory structure easy on the eyes, but it can be beneficial for other reasons. It ensures the Mac and iOS apps are always made ”cleanly”. Future Xcode releases may change aspects of how it generates Safari extension projects (I believe this already had happened in the past), so avoiding ”hard-coding” the Xcode project as part of the extension’s codebase might avoid problems down the line.

I exported a Safari extension and tested it on Safari 16.3 and 16.4 on macOS 13.3 and iOS 16.3, and all functionality appears identical to that of the extension on Firefox and Chrome.
</br>
</br>

@rdev, @CamilleScholtz, @1998code, curious to hear your thoughts as you submitted the PRs that I mentioned.

### DIFF
--- a/Extensions/combined/manifest-safari.json
+++ b/Extensions/combined/manifest-safari.json
@@ -1,0 +1,36 @@
+{
+  "name": "__MSG_extensionName__",
+  "description": "__MSG_extensionDesc__",
+  "default_locale": "en",
+  "version": "3.0.0.8",
+  "manifest_version": 2,
+  "background": {
+    "scripts": ["ryd.background.js"],
+    "persistent": false
+  },
+  "icons": {
+    "48": "icons/icon48.png",
+    "128": "icons/icon128.png"
+  },
+  "permissions": [
+    "activeTab",
+    "*://*.youtube.com/*",
+    "storage",
+    "*://returnyoutubedislikeapi.com/*"
+  ],
+  "browser_action": {
+    "default_popup": "popup.html"
+  },
+  "content_scripts": [
+    {
+      "matches": ["*://*.youtube.com/*"],
+      "exclude_matches": ["*://*.music.youtube.com/*"],
+      "run_at": "document_idle",
+      "css": ["content-style.css"],
+      "js": ["ryd.content-script.js"]
+    }
+  ],
+  "options_ui": {
+    "page": "popup.html"
+  }
+}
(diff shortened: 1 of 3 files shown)


1998code:
Hi @suhailskhan, just tried your code. Looks good to me.

⚠️ Few problems were found.
1) The numbers are repeated.
CleanShot 2023-03-31 at 00 03 00@2x
2) ```npm run dev:safari``` will be stuck in the terminal.
CleanShot 2023-03-31 at 00 08 33@2x
3) Maybe I am in macOS beta and got this warning. The extension still works fine.
CleanShot 2023-03-31 at 00 03 33@2x

✅ Checked:
1) Run ```npm run build:safari``` can build dist and open Xcode.
2) After signing, the extension can display correctly.
CleanShot 2023-03-31 at 00 02 25@2x

Thanks.

1998code (approved):

suhailskhan:
Thanks for the review, @1998code.
</br>

> 1. The numbers are repeated.

It looks like this is an issue stemming from the last upstream commit. Here’s Chrome when using the extension built from `main`:
image

> 2\. `npm run dev:safari` will be stuck in the terminal.

I see. I overlooked that the file watcher spawned by `npm run dev` would keep any subsequent commands from running, including the Safari extension generator. `npm run dev:safari` would probably not be necessary since Xcode and Safari themselves would provide the debugging workflow anyway.

> 3) Maybe I am in macOS beta and got this warning. The extension still works fine.

Looks like it’s a beta-related issue. Here is my output:
```
App Name: Return YouTube Dislike
App Bundle Identifier: com.returnyoutubedislike.safari-ext
Platform: All
Language: Swift
```

Anarios:
Wow, perfectly done.

PR #937, opened 2023-09-06
🌲 Swedish README

AminoffZ:
Added a Swedish README following the format.

I think it's decent enough (Swedish native speaker). Not sure I love using "dislikes" in Swedish, if anyone can think of terms that fit better and don't sound pedantic LMK. I feel like it's ok to sacrifice some level of faithfulness to a language in order to be understood.


Anarios (approved):
Thanks

Issue #735, opened 2022-08-26
Make It Work With Youtube Watch Page Fix Or Youtube Proper Description Script


TheBSODAndWSODFan:
### Extension or Userscript?

Extension

### Request or suggest a new feature!

https://youtu.be/N1h5ARS0uEw

When I Switching Account, The Like/DIslike Bar Is Missing When Youtube Watch Page FIx Script is enabled. That's Because My Second Account Has New Youtube Layout And I Personally Don't Like New Layout. And Enhancer For Youtube Extension Is Odd

Browser: Chromium Version 104.0.5112.101 (Official Build) for Linux Mint (64-bit)
RYD Extension Version: 3.0.0.5
Youtube Watch Page Fix Script: https://greasyfork.org/en/scripts/445015-youtube-watch-page-fix Or Youtube Proper Description Script: https://greasyfork.org/en/scripts/440613-youtube-proper-description

Extension Enabled: 
Improve YouTube 3.9.46
enhanced-h264ify 2.1.0
Enhancer for YouTube 2.0.115.1
Return YouTube Dislike 3.0.0.5
Tampermonkey 4.16.1
YouTube Redux 2.13.3

Sorry For Bad English And Sorry If This Issue Is Duplicate.

### Ways to implement this!

Add A Toggle Switch For Make It Work With Youtube Watch Page Fix Or Youtube Proper Description Script (Disabled By Default)

### Can you work on this?

- [ ] Yes
- [ ] No

### Will you be available for follow-up questions to help developers implement this?

Yes

JunkiEDM:
I use the `Youtube Proper Description` userscript and found that adding 
```
if (document.getElementsByTagName('ytd-watch-metadata').length > 0) {
    document.getElementsByTagName('ytd-watch-metadata')[0].remove()
}
```
at the end of the async function at the bottom of the script (screenshot below) fixes the issue for me. Haven't tested it too thoroughly but doesn't seem have broken anything.
(devs: yes i know this should be using variables and could be made cleaner but its just a temporary fix so i couldnt be bothered)

<details><summary>screenshot</summary>

![image](https://github.com/Anarios/return-youtube-dislike/assets/8350274/f5064524-b960-40a6-9f2e-3eb09ac70079)
</details>

[...]

PR #952, opened 2023-10-14
New YouTube UI fix for the latest version 3.0.0.11

tdreams2:
I changed this line of code in createDislikeTextContainer() function:
`const textNodeClone = (getLikeButton().querySelector("button > div[class*='cbox']") || getLikeButton().querySelector('div > span[role="text"]').parentNode).cloneNode(true);`

changed to this:
`const textNodeClone = (getLikeButton() || getLikeButton().parentNode).cloneNode(true);`

Selectors: ("button > div[class*='cbox']") and ('div > span[role="text"]') can't be found on the page

PR #959, opened 2023-10-16
Fix incompatibilities with some userscripts

JunkiEDM:
Fixes compatibility with scripts that revert the description layout such as ['Youtube Watch Page Fix'](https://greasyfork.org/en/scripts/445015-youtube-watch-page-fix) and ['Youtube Proper Description'](https://greasyfork.org/en/scripts/440613-youtube-proper-description), as described in issue #735.
Tested on desktop & mobile and it seems to be working all the same without those scripts, but let me know if you find anything misbehaving.


AminoffZ:
Check out #952, the UserScript there seems to be working, as does the chrome and firefox extension (with limited testing).

PR #992, opened 2023-11-30
Userscript fix for the latest YouTube update

Moondarker:
This should be backwards compatible if this is yet another A/B test from YouTube

Tested on: Windows 10, Chrome Stable 119.0.6045.161


Anarios:
Thanks

PR #1016, opened 2023-12-13
Implemented centralized versioning

Moondarker:
Version numbers were all over the place - different versions in localization files (`3.0.0.13`, and `3.0.0.1` in PL), GitHub manifest (`3.0.0.12`), extension stores manifest (`3.0.0.14`), and `package.json` (`2.0.0`).

This pull request aims to fix the problem by changing all the version references except for one in `package.json` to template strings. These template strings will be replaced to the actual build number from `package.json` at build time.

Because versioning scheme of this project is not compatible with SemVer, it looks different in `package.json` (e.g. `3.0.0-14` instead of `3.0.0.14`), but it's converted back to the currently used versioning scheme during the build.

### DIFF
--- a/Extensions/combined/_locales/cs/messages.json
+++ b/Extensions/combined/_locales/cs/messages.json
@@ -42,7 +42,7 @@
     "message": "aktualizovat na"
   },
   "version30installed": {
-    "message": "Verze 3.0.0.13 nainstalována"
+    "message": "Verze __RYD_VERSION__ nainstalována"
   },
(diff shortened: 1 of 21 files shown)


Moondarker:
Rebased to [b133ade](https://github.com/Anarios/return-youtube-dislike/commit/b133adeecbe6e7fb321c615c8ac9727466f6103c), @Anarios

Anarios:
Thanks a lot.

Issue #1037, opened 2024-02-23
New Youtube Format Bug


n0753w:
### Have you tried to find similar issues (open or recently closed)?

- [X] Yes, this issue is not a duplicate

### Browser

MS Edge

### Browser Version

Version 121.0.2277.128 

### Extension or Userscript?

Extension

### Extension/Userscript Version

v3.0.0.14

### Video link where you see the problem

Any Youtube video

### What happened?

The dislike counter doesn't show with the new ugly UI. Is it possible to make a workaround or does this addon only work with the "normal" UI?
![image](https://github.com/Anarios/return-youtube-dislike/assets/138560355/6992c503-8774-4dd5-9d23-f1bca81ffa82)

### How to reproduce/recreate?

This only happens because of the new UI. If someone else has said UI, then most addons will fail to work. 

### Will you be available for follow-up questions to help developers diagnose & fix the issue?

Yes

FilipLac:
I also have the same issue.

cocoKoniQ:
Can confirm, I have the same issue.

CobaltM:
yep same issue

greatwolf:
Same issue here.

CHlLL7:
Had the same issue with the new UI but as a workaround: Installing "Enhancer for Youtube" brings back the old UI and Return Youtube dislikes will work again

n0753w:
> CHlLL7

Yeah I also have that extension. It's great. About a week or so ago, the guy who made the YT Enhancer made an update to prevent the ugly ui from occurring, which is great.

Couchy:
Is this project dead? I could fix the UI, but there's not much point if the backend is also unmaintained.

n0753w:
Everything is back to normal for now on my side at least. YT likes to dish out experimental formats randomly.

RocketRobz:
The new format has rolled out recently on my side, and using "Enhancer for Youtube" also restores the dislikes on my side.

theblackhole:
This issue has been fixed with the latest release ([v3.0.0.16](https://github.com/Anarios/return-youtube-dislike/releases/tag/v3.0.0.16)) which is already available for Edge/Chrome in the Chrome Webstore (Firefox soon)

With this new release, there's also a minor issue where the buttons are aligned to the right when the extension is enabled. I provided a fix (#1047) and it should be available in a future update.

Orevin:
It looks like this issue resurfaced (extension 3.0.0.16)

[...]

PR #1047, opened 2024-04-19
Extension - fix spring 2024 UI buttons position

theblackhole:
With the new (spring 2024) UI, action buttons are aligned to the left but an obsolete patch made them align to the right instead.

This commit removes this obsolete patch in order to restore the original alignment (left for new UI and no changes/right for the previous UI).

Tested with the new and previous UI (Firefox and Edge on Windows) with and without the ratio bar

Screenshot of the issue:
[](https://github.com/Anarios/return-youtube-dislike/assets/7818904/7c0378fc-e168-4b1c-901d-55e264218c8b)

What it's supposed to look like (RYD extension disabled) :
[](https://github.com/Anarios/return-youtube-dislike/assets/7818904/b586f265-8433-4ed4-a691-56ded481880c)

Screenshot of the provided fix:
[](https://github.com/Anarios/return-youtube-dislike/assets/7818904/71db7120-a20e-42fd-8a82-c8426b496d6f)

### DIFF
--- a/Extensions/combined/src/bar.js
+++ b/Extensions/combined/src/bar.js
@@ -95,13 +95,6 @@ function createRateBar(likes, dislikes) {
           descriptionAndActionsElement.style.borderBottom =
             "1px solid var(--yt-spec-10-percent-layer)";
           descriptionAndActionsElement.style.paddingBottom = "10px";
-
-          // Fix like/dislike ratio bar offset in new UI
-          document.getElementById("actions-inner").style.width = "revert";
-          if (isRoundedDesign()) {
-            document.getElementById("actions").style.flexDirection =
-              "row-reverse";
-          }
         }
       } else {
         document.querySelector(`.ryd-tooltip`).style.width = widthPx + "px";


Anarios:
@theblackhole do you still experience the problem, is this patch still required?

I'm hesitant about removing older code - because many people can still have that version of UI.

theblackhole:
@Anarios Unfortunately I don't have the new (spring 2024) UI anymore so I can't tell if the patch is still required. But the current UI (before this big layout change) isn't affected by this patch, I think this was for an even older version of the UI

leakedmemory:
@Anarios it's not necessary anymore, I guess. Tested on version 3.0.0.14 with LibreWolf 127.0-1 and everything looks fine.

theblackhole:
@leakedmemory In your screenshot, return-youtube-dislike doesn't work at all 😅

leakedmemory:
@theblackhole Indeed, that was when I thought the bug of #1037 was not interfering with this, but I'm wrong. This still happens and that other bug to, so I'm gonna remove my other comment. Thanks.

Anarios:
Closing this one since it doesn't seem like anyone has a problem with this at the moment

PR #1059, opened 2024-06-14
Rename FAQpt_BRmd to FAQpt_BR.md

GhostifyYT:
FAQ for Brazillian Portuguese file was missing an extension dot causing it to display incorrectly.


Anarios:
Thanks

PR #1061, opened 2024-06-30
Localize the extension to Arabic

EyadMahm0ud:
Notes:

- Right-to-Left direction is not supported in the extenstion, not a big deal anyway.
- There was a missing `[` in CONTRIBUTINGfr.md.
- Had to add `<p dir="rtl">`, because GitHub doesn't care.


Anarios:
#1112 got merged earlier than this, which one is a better quality translation?

PR #1090, opened 2024-09-27
Add localization to the tooltip-porcentage-mode fields with the Spanish translation

jestra52:


jestra52:
The translation is a little bit longer than the English text, so it looks a little bit off:

imagen

Please let me know if that's ok

PR #1092, opened 2024-10-08
Fix links to the Chrome Web Store extension page

jake-anto:
Apparently, the trailing slash at the end of the URL breaks the link. Instead of just removing the trailing slash, I updated the links to point to the final redirected page.


Anarios:
Thanks a lot

MrQS94:
The link is down right?

PR #1096, opened 2024-10-15
Unobserved child node removals

KibbeWater:
Due to a recent youtube update, like/dislike buttons may re-render causing "foreign" elements to get cleared


KibbeWater:
This will also fix #1095

Anarios:
Thanks

PR #1144, opened 2025-06-13
Improve the German README

dtdan-03:
- Corrected spelling and grammar
- Improved wording and sentence structures
- make it sound more natural

(Hint: "Rate limit" stays in English as an anglicism → the technical term must not be translated here imho)


Anarios (approved):
Thanks a lot

Issue #1153, opened 2025-07-09
pressing like causes infinite async loop of solvepuzzle method


8bit-meiko:
### Have you tried to find similar issues (open or recently closed)?

- [x] Yes, this issue is not a duplicate

### Browser

firefox

### Browser Version

138.0.4

### Extension or Userscript?

Extension

### Extension/Userscript Version

3.0.0.18

### Video link where you see the problem

https://www.youtube.com/watch?v=8X2dL1d_C8s

### What happened?

when i press like on any uploaded video, the plugin performs the solvepuzzle method, but it infinite loops and causes high single thread cpu usage as it does not stop by itself. 

attached are 2 screenshots of my about:performance breakdown whats happening.

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/23729d74-6a3c-4d61-80ea-149b828351af" />
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/51ec17e1-0a48-49c5-b4e5-5528a7218ff3" />

initially i got fooled by thinking the extention has a cryptominer baked in (blaming myself for jumping to conclusions and not diving into the root cause)

what should be happening: the solvepuzzle method to not repeat infinitely.

### How to reproduce/recreate?

press like on any video, see the extention's cpu usage go up on a single thread by the infinite loop.

other extentions that i run (affecting youtube) for transparency: ublock origin, dark reader and nuxify.

### Will you be available for follow-up questions to help developers diagnose & fix the issue?

Yes

KibbeWater:
Exactly what are your steps when this issue occurs, does it still occur? do you do any specific actions after navigating to the page?

What exactly are you measuring when you show the performance graphs? There are loads of options to measure from in about:performance.

KibbeWater:
Can't seem to repro on my own installation.

I'm curious if maybe it's cause of the challenge input returned from the server, could you do the following steps and run a script to check what is being sent from the server?

1. Enter the firefox extension debug page `about:debugging#/runtime/this-firefox`
2. Click inspect on "Return YouTube Dislike"
3. In the new window, enter the console and paste the script from below
4. Take a screenshot of the output

```js
(async () => {
	console.clear();
	console.log('Start logging');

	const { userId } = (await browser.storage.sync.get(null)) ?? {};
	if (!userId) return console.log('No UserID has been saved/generated');

	const voteResponse = await fetch(`https://returnyoutubedislikeapi.com/interact/vote`, {
[...]

PR #1155, opened 2025-07-12
Infinite register/vote loop on sendVote returning 401

KibbeWater:
When a registered user receives a 401 upon voting, it will generate a new userId and register it. 
However, if the server repeatedly returns 401 for the same user every time, the user will get stuck in a voting loop that will spam request to the /vote endpoint and spike the users CPU usage on running solvePuzzle

This change adds a safeguard against this infinite loop edgecase and prevents liking a video to spike the users CPU

### DIFF
--- a/Extensions/combined/ryd.background.js
+++ b/Extensions/combined/ryd.background.js
@@ -96,7 +96,7 @@ api.runtime.onInstalled.addListener((details) => {
 //   }
 // });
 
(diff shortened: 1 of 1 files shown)


KibbeWater:
This PR closes #1153

Anarios (approved):
Thanks

Anarios:
I wonder why register fails though.

Issue #1274, opened 2026-04-20
Not working anymore right now


sham-pain:
### Have you tried to find similar issues (open or recently closed)?

- [x] Yes, this issue is not a duplicate

### Browser

Firefox

### Browser Version

149.0.2

### Extension or Userscript?

Extension

### Extension/Userscript Version

3.0.0.18

### Video link where you see the problem

https://www.youtube.com/watch?v=baApwsg9PYA

### What happened?

I don't use Youtube every day but as of today I found that dislikes don't show anymore with the extension on. The link is just an example I'm listening to right now, but it's on all videos for me. I went to private mode and turned off other extensions as well.

### How to reproduce/recreate?

Open a Youtube video and look at likes/dislikes - only the number of likes (thumbs up) show.

### Will you be available for follow-up questions to help developers diagnose & fix the issue?

Yes

Teslahaha:
Confirmed

Anarios:
Looks like a new UI on YouTube being tested. Shouldn't be a problem

Aragorn-E:
Yeah, I can confirm that it's broken as well.

Version:

<img width="330" height="426" alt="Image" src="https://github.com/user-attachments/assets/94cb3201-83ad-4092-99f7-4f10fe55c746" />

Mouadabdelghafouraitali:
Yes, but the API is still working, it’s just the extension that isn’t working.

Anarios:
The quick and dirty fix will look like this:

the icon is lost - but at least we will have the count. 

The benefit of this fix is that it can be immediate - i.e. backend only - without wait time for the extension update approval.

<img width="415" height="115" alt="Image" src="https://github.com/user-attachments/assets/f8adbef8-daae-4c43-939a-6fea129a8b03" />

Anarios:
The dirty fix is out, the full fix time will depend on the reviewers at chrome\mozilla stores

Mouadabdelghafouraitali:
> The dirty fix is out, the full fix time will depend on the reviewers at chrome\mozilla stores

Thank you for the hard work.

18209jg1:
Will the FULL fix time show the thumbs down icon?

18209jg1:
That didn't answer my question.

user21t:
[...]

PR #1292, opened 2026-06-14
Fix dislike count, icon, and ratio bar on new YouTube UI (#1274)

yaldar:
## Problem

YouTube's new watch-page UI (issue #1274) broke the extension:

- The dislike **count rendered without its thumbs-down icon** (or as a duplicated `00`).
- The like/dislike **ratio bar was invisible**.

Root causes, confirmed against the live new-UI DOM (`segmented-like-dislike-button-view-model` / `dislike-button-view-model` with camelCase `ytSpec*` classes):

1. The new `<like-button-view-model>` / `<dislike-button-view-model>` are inline wrappers, so `getComputedStyle(...).width` resolves to `auto` → `NaN`, breaking the ratio-bar width calculation.
2. The new UI **dropped the `--yt-spec-text-primary` / `--yt-spec-icon-disabled` design tokens**, so the bar's `background: var(--token)` (no fallback) collapsed to transparent — the bar was present and correctly proportioned, just invisible.
3. The live server-side selector "dirty fix" writes the count straight into the icon button via `innerText`, which wipes the icon and can leave a stray duplicate number next to our own container.

## Changes

- **`buttons.js`** — `removeStrayButtonText()` strips any count written directly into the dislike `<button>`, so the icon is preserved and the count isn't duplicated.
- **`utils.js` / `bar.js`** — `getElementWidth()` falls back to `getBoundingClientRect().width` when computed width is `auto`/`NaN`, fixing ratio-bar sizing for the inline view-model wrappers.
- **`content-style.css`** — theme-aware fallback colors for `#ryd-bar` / `#ryd-bar-container`; when YouTube's tokens exist, `var()` still prefers them automatically.
- **Tests** — new-UI regression test built from the real new-UI markup, a dirty-fix cleanup test, and `getElementWidth` unit tests.

## Testing

- `npm test` — 143 tests pass.
- `npm run build` — clean.
- Manually verified in Firefox (temporary add-on) on the new UI: the dislike count renders **with** its icon, and the ratio bar is visible with correct proportions (e.g. `3,513 / 191` → ~94.8% like share).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Extensions/combined/content-style.css
+++ b/Extensions/combined/content-style.css
@@ -18,15 +18,28 @@
   --yt-spec-brand-background-secondary: #000 !important;
 } */
 
+/* Fallback colors keep the ratio bar visible when YouTube's design tokens are
+   unavailable. The newer YouTube UI dropped these --yt-spec-* variables, which
+   left the bar fully transparent (invisible) even though its proportions were
+   correct. Light-theme fallbacks are defined here; dark-theme below. When the
+   tokens do exist, var() uses them and these fallbacks are ignored. */
 #ryd-bar-container {
-  background: var(--yt-spec-icon-disabled);
+  background: var(--yt-spec-icon-disabled, #909090);
   border-radius: 2px;
 }
 
 #ryd-bar {
-  background: var(--yt-spec-text-primary);
+  background: var(--yt-spec-text-primary, #0f0f0f);
   border-radius: 2px;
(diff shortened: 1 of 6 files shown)


yaldar:
I created this PR with the help of Claude app (to be honest Claude did all the work) but it works on my machine.

PR #1302, opened 2026-08-05
Generate the userscript from a shared codebase instead of hand-maintaining it

jonathansampson:
## Summary

The userscript (`Return Youtube Dislike.user.js`) and the browser extension implemented the same selectors, DOM handling, and formatting logic twice, by hand, in separate files. Every fix had to be applied in both places to stay in sync, and nothing enforced that. This PR extracts the shared logic into a common module and adds a build pipeline that generates the userscript from it, so there's one source of truth.

## What changed

**Shared codebase**
- Moved `state.js`, `utils.js`, `buttons.js`, `bar.js`, `config.js`, `events.js` (and tests) out of `Extensions/combined/src` into a new `Extensions/common`, leaving the extension-only code (chrome storage, Patreon OAuth, premium analytics UI, manifests) behind in `Extensions/combined`.
- Split `state.js` further into dependency-free leaf modules (`dom.js`, `runtime-state.js`) to break a circular dependency between `state.js`/`buttons.js`/`bar.js`/`utils.js` that Rollup's stricter analysis flagged.
- Cleaned up dead code and exports that had no callers outside their own module (`getState()`, several over-exported internals), found via webpack's tree-shaking report plus a manual audit.

**Userscript build pipeline**
- Added `Extensions/UserScript/src/userscript-entry.js`, which composes the same shared modules the extension uses, plus two small override files (`utils.userscript.js`, `bar.userscript.js`) for the userscript's storage shim and its two extension-less options (`numberDisplayRoundDown`, `rateBarEnabled`).
- Built with Rollup rather than webpack (which the extension keeps using) — Rollup flattens the module graph into one file with no runtime wrapper, matching the style of the original hand-written script instead of producing bundler boilerplate.
- `npm run build:userscript` writes straight to the existing committed path, so already-installed copies keep auto-updating from the same `@updateURL`. CI now rebuilds the userscript and fails the build if the committed file has drifted from source.
- Added a generated-file notice in the userscript header, and documented the `Extensions/common` split and the build step in `CONTRIBUTING.md`, so contributors don't hand-edit the generated file by mistake.

**Tests**
- Added coverage for the storage shim, the `utils.userscript.js` override (including a drift guard against `utils.js`'s actual exports), and a sanity check that the committed `.user.js` file has a well-formed metadata block and parses as valid JS.

**Housekeeping**
- Reformatted `Extensions/*.js` and then the whole project with Prettier per `CONTRIBUTING.md`'s formatting guideline (no functional changes).

No behavior change to either the extension or the userscript's runtime output — this is a restructuring of how the userscript is produced, not a change to what it does.


Anarios:
Thanks for putting this together—the core idea is good, and I appreciate the thought and effort that went into it. Unfortunately, the PR combines the userscript build approach with a broad set of refactors and related changes, making it too unfocused to review and merge safely in its current form.

I’m going to close this PR, but I plan to reuse the general idea in my own, more focused implementation for the userscript. Thanks again for the contribution and for exploring this direction.

